PR #1074, opened 2015-01-03
Use HTML meta tag and HTTP header to avoid leaking info through referrer

### LINKED ISSUE #1068: make use of html meta referer directive set to never
As suggested by @diracdeltas on https://github.com/freedomofpress/securedrop/issues/819 for securedrop, it could be interesting to use the html meta referrer set to never also on globaleaks in order to limit the possibility for referrer leakage.

surely this won't be honoured by all browsers, nor from the latest ones but it surely limits somehow more what happens. in additions there would be no problem to add it on the application as the clientside logic currently do not make use of referer info for nothing.

mapmeld:
( new PR for devel branch, as requested by @evilaliv3 )

Modifies server and client to fix #1068

HTML tag based on https://wiki.whatwg.org/wiki/Meta_referrer

HTTP header based on https://w3c.github.io/webappsec/specs/content-security-policy/#directive-referrer

### DIFF
--- a/backend/globaleaks/handlers/base.py
+++ b/backend/globaleaks/handlers/base.py
@@ -220,6 +220,9 @@ class BaseHandler(RequestHandler):
         self.set_header("Pragma", "no-cache")
         self.set_header("Expires", "-1")
 
+        # to avoid information leakage via referrer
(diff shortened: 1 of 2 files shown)


evilaliv3:
well done @mapmeld ! goin to merge this!

Issue #1267, opened 2015-05-10
Make it easier delete multiple Tips for a receiver


vecna:
In case of congestion, tips need to be deleted without clicking on every button from the outside, and without having a warning every time is clicked on it, and without having a DELETE command for every Tip. Still, thinking to the open data feed + end2end encryption change seriously this issue. so before think to the steps proposed, is better if we analyze the doubts below.

### Steps

  * delete button has to be moved in another point of the page
  * Receiver "select" the tip in order to apply the command later, like gmail with the delete
  * REST API taking a list of Tip ID

### Doubts 
 
I don't think this procedure can solve a flood in the end2end model. The preview will not be usable anymore, the user will require the possibility to tag/name a Tip like #1156 and for sure an user will access to **all** the tip before delete them. 

In addition, the delete concept maybe should be forget. the "action" that a receiver express is "this tip is meaningful to be analyzed" or "this tip is not meaningful", the effect for the receiver will be the same, but for the open data feed this will correspond more truthfully to "10 submission received, 90% shit" and not to "10 submission received, 20% shit" (with 70% deleted, not disclosed).

evilaliv3:
i'm ok with what specified here!

i suggest to implemente the DELET in the TipsCollection handler that already exists and has only a get().
the delete would take an array of the tips id were an empty array can be used to mean "all tips".
an like in gmail/wordpress/whatever we can have the classical select box on the top of the table were to have: checkall/uncheck all: bulkaction:postpone/delete

vecna:
And the Delete event has to be recorded in the stats then, just I don't know where yet....

fpietrosanti:
It may have relationship with #1224 ?

vecna:
@fpietrosanti yes, but only because I'm thinking to split the tip lists (new, tagged, "new-name-for-deleted-tips") and eventually we can implement paging (that will address completely the issue you are speaking of)

fpietrosanti:
Regarding paging there's a tons of libraries and approaches to do that with AngularJS, don't know the right approach client-side and the impact/requirements for search/selectors server-side:
http://ngmodules.org/modules/Angular-Paging
https://github.com/mizrael/AngularJs-Pagination

vecna:
## Sequence of the operations: 

  * handlers review, paging at handler level, unit test in that side
  * tip interface (buttons, numbers, minor enhancement and #1156 #1260)
  * landing page receiver (number of tip per category) 
  
[...]

PR #1315, opened 2015-06-24
Receiver can delete (or postpone) multiple tip at once -- Issue/1267

vecna:
From D8.6 issue #1267


evilaliv3:
i've few things that i want to change like the direct usage of jquery and the abuse of rootscope but in general it do what it says to do! well done bro'!

PR #1342, opened 2015-07-29
Disk switch + Tip improvement + new texts

vecna:
Note: the new text has to be reviewed (and sent to Transiflex when reviewed)

The idea is to:

  * separate the technical info from the submission info
  * put a small hint that explain, to every receiver "you can develop a better Questionnaire"
  * put the word Questionnaire inside the interface
  * TODO: we need a proper stats (5% 95% is made up based on TI + AWP feedback)


fpietrosanti:
Receivers actually does not handle/manage questionnaires, it's a planned and requested features, but not yet available.

vecna:
yes @fpietrosanti I'm not saying that. I'm saying that receiver will see, separately, technical info from submission info. and above the submission info, an hint that make them think out to more complex questionairre. then they have to ask to the admin. but is from that persons that the request has to become. (otherwise, as we saw, they don't see the potentiality)

fpietrosanti:
@vecna not all users may accept that visualization, for example not in any large organization. 
So i'd suggest to:
- make the hint configurable (enable/disable)
- make it be visible only if the questionnaires items on the tip are "less than X" (because it doesn't make sense to visualize this tip on ALAC or ASLBrescia)

vecna:
just review the text, can be display when the node is keeping the default fields

evilaliv3:
ok patch revised and it sems all good to me aside from the Questionarie Suggstion.

in relation to the "Questionaire Suggestion" i'm ok for such a need but not with the implementation so that i'm going to integrate the patch without it and opening a ticket for such a feature and archiving there the texts already written.

The reason is the same as @fpietrosanti. we can't forcefully add such a thing so that when we implement features like this we would need:
- a possibility to disable it
- be sure that the text is shown only when needed. e.g. right know that suggestion should be shown only if one is using default fields but: 1) an admin may want to use default and not suggest to change them to receivers; 2) receivers are not in possibility to change them, the text provided is not actionable 3) technically we are not in possibility to detect if one is using default fields with a check 100% valid. 

as stated this

backend/globaleaks/handlers/submission.py
```diff
@@ -264,11 +268,9 @@ def post(self):
...
         Errors: ContextIdNotFound, InvalidInputFormat, SubmissionFailFields
 
-        This creates an empty submission for the requested context,
-        and returns submissionStatus with empty fields and a Submission Unique String,
-        This is the unique token used during the submission procedure.
-
-        This create a Token, require to complete the submission later.
+        This API create a Token, a temporary memory only object able to keep track of
+        the submission. If the box is under stress, complete the submission will require
+        some effort (hashcash, captcha) kept track in the Token.
```
evilaliv3: maybe the box should be called always system to use something that may recall to a computer?

client/app/styles/main.css
```diff
@@ -44,7 +44,8 @@ th {
   color: #FFF;
   background: #428bca;
   border-bottom: 1px solid #357ebd;
-  padding: 30px 15px 30px 15px;
+  padding: 5px 15px 5px 15px;
+  border-radius: 2%;
```
evilaliv3: border-radius on the header, hu?!

client/app/styles/tip.css
```diff
@@ -1,6 +1,7 @@
 .TipPageSectionTitle {
-  font-size: 20px;
-  font-weight: bold;
+  font-size: 17px;
+  font-weight: bolder;
+  font-variant: small-caps;
```
evilaliv3: nice change the 'small-caps' but lets first evaluate if it's ok to use it for every title all along the application and keep things uniform to not create confusion.

i'm also thinking to find some good product and copying its style as a whole :)

client/app/styles/tip.css
```diff
@@ -23,3 +24,36 @@
 .fieldAnswer {
   margin-bottom: 10px;
 }
+
+.SubmissionInfoSection {
+  background-color: #eaebe2;
+  border-radius: 2%;
```
evilaliv3: in relation to borders never use percentage as its obviously a related to the size of the box so that in the end every box will have fancy different style

client/app/views/receiver/tip.html
```diff
@@ -79,19 +79,65 @@
...
+
+
+        </tbody>
+      </table>
+      </div>
+
+      <hr />
+
+      <div class="TipPageSectionTitle" data-translate>The filled Questionnaire</div>
+      <small>
```
evilaliv3: let's try always to avoid the usage of small/b/i/and ather graphic modifier as it will make difficult for adopters to customize them with css; assign css classes and try to reuse them along the application.

http://www.w3.org/TR/html401/present/graphics.html

"Although they are not all deprecated, their use is discouraged in favor of style sheets."

client/app/styles/tip.css
```diff
@@ -23,3 +24,36 @@
...
+}
+
+.SubmissionContentSection {
+  background-color: #e5fecd;
+  border-radius: 2%;
+}
+
+.left {
+  text-align: right;
```
evilaliv3: when one provides special classes for alignment, he should provide the same reversed in rtl.css

e.g. for the .left one should create a .left inside rtl.css with written "text-align: left" instead of "text-align:right"

client/app/styles/tip.css
```diff
@@ -23,3 +24,36 @@
...
+  padding: 0.5%;
+  width: 30%;
+}
+.right {
+  text-align: left;
+  padding: 0.5%;
+  width: 65%;
```
evilaliv3: we are using bootstrap grid system for a reason that is that it that makes things responsive. avoid using of percentages directly.

client/app/views/receiver/tip.html
```diff
@@ -79,19 +79,65 @@
...
+            {{::tip.expiration_date | date:'dd-MM-yyyy HH:mm'}}
+          </td>
+        </tr>
+        <tr>
+          <td class="left">
+            <span data-translate>Whistleblower security</span>
+          </td>
+          <td class="right">
+            <span data-ng-hide="tip.tor2web"
+                  data-translate class="btn-success">Anonymous Submission</span>
```
evilaliv3: the regexp that push strings on transifex expect data-translate to be immediatly befor the the closing >, e.g. ".... data-translate>"

client/app/views/receiver/tip.html
```diff
@@ -79,19 +79,65 @@
...
+            <span data-translate>Expiration Date</span>
+          </td>
+          <td class="right">
+            {{::tip.expiration_date | date:'dd-MM-yyyy HH:mm'}}
+          </td>
+        </tr>
+        <tr>
+          <td class="left">
+            <span data-translate>Whistleblower security</span>
+          </td>
```
evilaliv3: @fpietrosanti:  that do you think to this "Whistleblower security"? is there the possibility to write something more neutral given the fact that e.g. for the italian law a whistleblower is only who report things being an employee of a public agency?

I don't like Submission security but its for sure more neutral.

client/app/views/receiver/tip.html
```diff
@@ -79,19 +79,65 @@
   <div class="row">
     <div class="col-md-6">
       <div class="TipPageSectionTitle" data-translate>Submission</div>
-      <div class="col-md-12">
-        <dl>
-          <dt>
-            <span>{{::tip.context_name}}</span>
+
```
evilaliv3: all this changes applied to receiver/tip.html should be done to whistleblower/tip.html

be homogenous my friend! :)

i will take care of doing it by finalizing the pull request and i will try also to use partial templates whenever possible.

client/app/views/receiver/tip.html
fpietrosanti: I'd use a a non-textual representation with an icon somewhere with a icon representing a "Connection" that can then have a value: HTTPS / Tor .
Then a hint with on-mouse-over that will say:
HTTPS: "The connection to send this submission was encrypted using HTTPS, ensuring confidentiality"
Tor: "The connection to send this submission was encrypted using Tor, ensuring confidentiality and anonymity" 

Then this indicator must be configurable on/off (because it's a TECHNICAL indicator, that's not likely in the interests of any lawyer handling anticorruption submissions).

evilaliv3:
here we go; i've applied @vecna patch but implemented all using bootstrap and making possible to customize the template by css in all it's things that are now homogenous.

screenshot from 2015-08-03 01 06 56
screenshot from 2015-08-03 01 07 02
screenshot from 2015-08-03 01 13 03

Issue #1269, opened 2015-05-11
Research strategy and test for anti-DoS protection


vecna:
Some issue report about performance doubt ( #1224 ), some issue are remain open because due to lacking of impact analysis on the anti-DoS strategies ( #825 #797 ) and the load caused in a system.

So I was thinking to perform some structured tests, collect data, and provide the defaults for the anti-DoS in a meaningful way. with these data, we can also provide meaningful information to a third party reviewer concerned about delays, captchas and tokens.

In order to do so, I've to (code):

  * find out a python profiler 
  * re-enable Storm debug and check timings involved in the operation
  * test the proctractor and perform parallel execution 
  * makes dump timing information in JSON format
  * put all the "research" options as part of a dedicated option, like Storm debug at the moment is -s
  * maybe recover the stress tester based in Twisted, never updated with the current API.

And research:

  * test login exponential delay, add token to that and compare differencies 
  * test over submission
  * test over parallel submission with fields up to the limit
  * test with file upload in parallel

/cc @evilaliv3 @fpietrosanti

fpietrosanti:
@vecna Did we ever considered "performance" as an issue of the result of a DOS? The issues has been in an excessive amount of submissions to be managed from the receivers point of view and a problem with notifications, but not a performance problem. 
Btw from the implementation, because of the end-to-end investment on GLC, it would be much better to make any kind of flood-testing with GLC itself improving it without writing code that will become outdated in few months because used only for flood testing.

vecna:
@fpietrosanti I mean, is part of the anti-DoS understand (through profiling) why sometime DB keep seconds to perform answer. if that can be controlled or influenced from the outside, this will make all the activities performed via GL delayed (because the transact is a single, blocking, thread). So what affect some performances, has to be understand clearly also for the reliability point of view.

I agree with GLC consideration, I can just get rid of twisted stresser, but if proctractor do not permit me to control threads and parallel activities (one of the possible attack scenarios), I've to find a workaround, I don't want make a test suite overcomplicated and time expensive.

evilaliv3:
i agree that at the stage we are writing the test for parallel submission with protractor and with a possible configuration in the number of requests and the number of parallel requests would be valuable.
if you want to concentrate a bit on writing the code for the Storm debugging code in the meantime i would add this test (i would need not more of a day for finishing current fixed to unit testing).

in order to be able to analyze the possible openes from performances issues to DoS issues it would be important for the possibility to log:
[...]

PR #1357, opened 2015-08-21
Anti dos research

vecna:
This is not to merge, is kept PR just to benefit of the branch compare.


vecna:
Ok probably I'll just cherry-pick the meaningful stuff when the research is complete, and this branch will be deleted without merge.

vecna:
This is #1269

PR #1401, opened 2015-09-03
Anti Dos Improvement + ReceiverTip Labels

vecna:
@evilaliv3


vecna:
stats_usage is done only in stress testing, and every handler just flood with emails at every instance, that's why.

PR #1429, opened 2015-09-15
Issue/799

vecna:
* based on the .105 version
  * UX do not manage effectively the loading time
  * configurable difficulty changing the number of zeros in HASH_ENDS_WITH and isAGoodPOW() client side function
  * used openpgpjs to generate the sha256

### DIFF
--- a/client/app/scripts/controllers/submission.js
+++ b/client/app/scripts/controllers/submission.js
@@ -90,26 +90,10 @@ GLClient.controller('SubmissionCtrl',
     return false;
   }
 
-  var iterateOverSHA = function() {
-    /* temporarly this section of code is put in the countdown */
-    console.log("Hashcash bruteforce starts now");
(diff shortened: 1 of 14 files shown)


client/app/scripts/controllers/submission.js
```diff
@@ -69,9 +70,44 @@ GLClient.controller('SubmissionCtrl',
     $scope.contextsOrderPredicate = 'presentation_order';
   }
 
+  var isAGoodPOW = function(binaryhash) {
+
+    if (binaryhash.charCodeAt(31) == 0 /* && binaryhash.charCodeAt(30) == 0 */ ) {
```
evilaliv3: @vecna: i think that it was wrong this part commented right?

evilaliv3: ah ok got it!

Issue #622, opened 2013-09-10
Refactor the logging related code


hellais:
Currently the logging code is a bit sketchy and it could do with some love.

In particular calling flush every time could create some issues and it's also somewhat scary to have some cases where you do not write to the log.

The goal of this is to refactor the code inside the observer in globaleaks/utils.py.
##

vecna:
To be addressed in this refactor: #782, #734

vecna:
@evilaliv3 I'll switch here the conversation started in https://github.com/globaleaks/GlobaLeaks/commit/11b1bf7d219c1e1211d2b2f75d604b0bff039c3a#commitcomment-13338747

## New table proposed

  * id INT NOT NULL: Note, will be the first Incremental Numeric-only ID we put in the system. In addition, there are not reference. Looking inside, hopefully, will be a little faster.
  * code TEXT NOT NULL: This is the Log code, usable to associate a text (translated, too) to the log
  * args VARCHAR: optional args, is a JSON, the number of args are forced. so can be 0, 1, 2, ... 
  * log_date TEXT NOT NUL: This is the time when the event is happen (for example: "a new tip generated" or "a mail has not been delivered for a mailserver error" )
  * subject TEXT NOT NULL CHECK(type IN ('admin', 'receiver', 'itip')): Based on the kind of subject, we know who is going to see this log. Only three interface will be influenced by this log system:
      * The admin (that probably will have a dedicated page, plus, red balloon for warning, in the landing page)
      * The receiver, that will see in a dedicated page the last events associated, and red balloon for warning, in the landing page.
      * The internaltip, so, receiver(s) and WB, this will be display in a dedicated section of the tip page.
  * subject_id TEXT: the "_id" is not a reference because can contain different data, dependently from the **subject** above. so, if is a 'itip' subject, this will be an InternalTip.id, if is a 'receiver', is a Receiver.id, if is an Admin, at the moment is None, when multiple Admin shall be supported, an Admin.id (or User.id, we'll see) 
  * repeated INT: if a log get repeated, like, an email is gone badly delivered, and this happen for all the emails, then is pointless report 4 or 20 same event. this counter get incremented **every time the last event get repeated - only if the args are the same**, if the args differ, is reported a new entry. is like the syslogd optimization,
  * last_repetition_date TEXT: this is the date of the last repetition, if a log keep going repeated, you'll see "Xth of event" and the last at last_repetition_date (the first, as log_date)

[...]

Issue #734, opened 2013-12-13
Improve security with automatic attack detection


fpietrosanti:
It has been noted that in a typical globaleaks deployment there's no specific application support to detect and report attacks.

This ticket is about providing support in globaleaks for:
- Network Intrusion Detection
- Host Intrusion Detection
- Log Monitoring
- Alerting (if an anomaly is triggered)

Given the globaleaks use-case, all of that kind of functionalities must not add any additional complexity to the end-user and must automatically be configured.

To do so we could use as a baseline the "harden-nids" package of ubuntu:
https://apps.ubuntu.com/cat/applications/precise/harden-nids/
http://packages.ubuntu.com/lucid/harden-nids

This ticket is about integrating harden-nids into globaleaks, in a way that does not require any end-user choice at installation time and not a single manual configuration.

PR #1435, opened 2015-10-05
Log refactor

vecna:
#622 related code

### DIFF
--- a/backend/globaleaks/db/sqlite.sql
+++ b/backend/globaleaks/db/sqlite.sql
@@ -437,7 +437,7 @@ CREATE TABLE log (
     args VARCHAR,
     log_date TEXT NOT NULL,
     subject TEXT NOT NULL,
-    subject_id TEXT,
+    log_message TEXT NOT NULL,
     log_level INTEGER,
     mail INTEGER,
     mail_sent INTEGER,
--- a/backend/globaleaks/db/sqlite.sql
+++ b/backend/globaleaks/db/sqlite.sql
@@ -441,8 +441,6 @@ CREATE TABLE log (
     log_level INTEGER,
     mail INTEGER,
     mail_sent INTEGER,
-    repeated INTEGER,
-    last_repetition_date TEXT,
     UNIQUE (id),
     PRIMARY KEY (id)
 );
--- a/backend/globaleaks/db/sqlite.sql
+++ b/backend/globaleaks/db/sqlite.sql
@@ -429,16 +429,20 @@ CREATE TABLE archivedschema (
     schema BLOB NOT NULL,
     UNIQUE (hash, type, language),
     PRIMARY KEY (id)
-)
+);
 
 CREATE TABLE log (
-  id INT NOT NULL,
-  code TEXT NOT NULL,
-  args VARCHAR,
-  log_date TEXT NOT NULL,
-  subject TEXT NOT NULL CHECK(type IN ('admin', 'receiver', 'itip')),
-  subject_id TEXT,
-  repeated INT,
-  last_repetition_date TEXT,
-  PRIMARY KEY (id)
-)
+    id INTEGER NOT NULL,
+    code TEXT NOT NULL,
+    args VARCHAR,
+    log_date TEXT NOT NULL,
+    subject TEXT NOT NULL CHECK(subject IN ('admin', 'receiver', 'itip')),
+    subject_id TEXT,
+    log_level INTEGER,
+    mail INTEGER,
+    mail_sent INTEGER,
+    repeated INTEGER,
+    last_repetition_date TEXT,
+    UNIQUE (id),
+    PRIMARY KEY (id)
+);
--- a/backend/globaleaks/db/sqlite.sql
+++ b/backend/globaleaks/db/sqlite.sql
@@ -436,7 +436,7 @@ CREATE TABLE log (
     code TEXT NOT NULL,
     args VARCHAR,
     log_date TEXT NOT NULL,
-    subject TEXT NOT NULL CHECK(subject IN ('admin', 'receiver', 'itip')),
+    subject TEXT NOT NULL,
     subject_id TEXT,
     log_level INTEGER,
     mail INTEGER,
--- a/backend/globaleaks/db/sqlite.sql
+++ b/backend/globaleaks/db/sqlite.sql
@@ -430,3 +430,15 @@ CREATE TABLE archivedschema (
     UNIQUE (hash, type, language),
     PRIMARY KEY (id)
 )
+
+CREATE TABLE log (
+  id INT NOT NULL,
+  code TEXT NOT NULL,
+  args VARCHAR,
+  log_date TEXT NOT NULL,
+  subject TEXT NOT NULL CHECK(type IN ('admin', 'receiver', 'itip')),
+  subject_id TEXT,
+  repeated INT,
+  last_repetition_date TEXT,
+  PRIMARY KEY (id)
+)
--- a/backend/globaleaks/jobs/log_sched.py
+++ b/backend/globaleaks/jobs/log_sched.py
@@ -5,13 +5,12 @@
 #
 import os
 from twisted.internet.defer import inlineCallbacks, returnValue
-from storm.expr import Desc, Asc
+from storm.expr import Desc
 
-from globaleaks.anomaly import Alarm, compute_activity_level
 from globaleaks.jobs.base import GLJob
-from globaleaks.settings import GLSettings, transact, transact_ro
-from globaleaks.models import Stats, Anomalies, Log
-from globaleaks.utils.utility import log, datetime_now, datetime_null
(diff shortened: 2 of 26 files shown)


evilaliv3:
@vecna: is this branch ready from your point of view and i should start reviewing it?

i see that some unit tests are still failing. can you please fix them?
(for what relates to the migration unit tests i could take care of it while implementing the migration but simply take some notes of the changes that relates to the db)

vecna:
This branch is not ready for the integration, because lack of the two sub-issue that can be at least pushed along ( #782, #734 ) but in this PR can be devel be aligned, review can start. review can start because the main concept are done, UT are failing, I'll address them, mostly we have to extend UT because the condition of caching are more than a couple.

vecna:
@evilaliv3 please start some review, I'll push soon the email notification system, but you'll see that the picklogs function can be reviewed.

backend/globaleaks/db/sqlite.sql
```diff
@@ -429,4 +429,18 @@ CREATE TABLE archivedschema (
...
+
+CREATE TABLE log (
+    id INTEGER NOT NULL,
+    code TEXT NOT NULL,
+    args VARCHAR,
+    log_date TEXT NOT NULL,
+    subject TEXT NOT NULL,
+    log_message TEXT NOT NULL,
+    log_level INTEGER,
+    mail INTEGER,
```
evilaliv3: missing NOT NULL

backend/globaleaks/db/sqlite.sql
```diff
@@ -429,4 +429,18 @@ CREATE TABLE archivedschema (
...
+);
+
+CREATE TABLE log (
+    id INTEGER NOT NULL,
+    code TEXT NOT NULL,
+    args VARCHAR,
+    log_date TEXT NOT NULL,
+    subject TEXT NOT NULL,
+    log_message TEXT NOT NULL,
+    log_level INTEGER,
```
evilaliv3: missing NOT NULL

backend/globaleaks/db/sqlite.sql
```diff
@@ -429,4 +429,18 @@ CREATE TABLE archivedschema (
     schema BLOB NOT NULL,
     UNIQUE (hash, type, language),
     PRIMARY KEY (id)
-)
+);
+
+CREATE TABLE log (
+    id INTEGER NOT NULL,
+    code TEXT NOT NULL,
+    args VARCHAR,
```
evilaliv3: missing NOT NULL

backend/globaleaks/db/sqlite.sql
```diff
@@ -429,4 +429,18 @@ CREATE TABLE archivedschema (
...
+    log_date TEXT NOT NULL,
+    subject TEXT NOT NULL,
+    log_message TEXT NOT NULL,
+    log_level INTEGER,
+    mail INTEGER,
+    mail_sent INTEGER,
```
evilaliv3: missing NOT NULL

backend/globaleaks/db/sqlite.sql
```diff
@@ -429,4 +429,18 @@ CREATE TABLE archivedschema (
...
+    UNIQUE (id),
```
evilaliv3: not needed, a primary key is always UNIQUE

backend/globaleaks/jobs/log_sched.py
```diff
@@ -0,0 +1,119 @@
...
+
+        max_id_check = store.find(Log)
+        max_id_check.order_by(Desc(Log.id))
+
+        if max_id_check.count() == 0:
+            log.debug("First initialization spotted!")
+            LogSchedule.highest_logged_id = 1
+        else:
+            LogSchedule.highest_logged_id = max_id_check[0].id
+
```
evilaliv3: the correct way to do this is simply: store.find(Log).max(Log.id)

check what is the  result when the table is empy and use it accordingly.

backend/globaleaks/jobs/log_sched.py
```diff
@@ -0,0 +1,119 @@
...
+    @transact
+    def dump_fresh_logs(self, store):
+        """
+        Look in the current memory storage of Logs, if there is
+        something not yet logged in the DB.
+
+        In the meanwhile, if the number of entries overcome the
+        50 elements, delete the old one. This mean that, worst case
+        scenario, we have in memory 50 Object x (Tip Numb + Receiver Num + Admin)
```
evilaliv3: this thing of dropping logs is wrong; for accountability reasons we cant never delete logs so that if we get the memory limits we have simply to write them on the disk and a single query of write won't weight more of a submsission so it's ok to do it automatically.

what i would suggest is:
- write when logs > memory_logs_limit
- have a scheduler that flush every minute anyhow (a single write won't weight more than all the queries performed during a submission)
- since now for reliabiliy reasons we need to handle shutdown signals (e.g.: CTRL + C)and perform flush of the data that is in memory before shutting down the app.

\cc @fpietrosanti

backend/globaleaks/jobs/log_sched.py
```diff
@@ -0,0 +1,119 @@
...
+    def clean_itip_log_in_memory(self, internaltip_id):
+        """
+        :param internaltip_id:
+
+        When an InternalTip get deleted, all the logs will remain in the
+        DB, but not in memory. This is the function to clean it.
+
+        NOTE: has print inside because the itip section has still to be tested
+        client side
+        """
```
evilaliv3: this comment is not clear.
what is the aim of this function?

backend/globaleaks/models/__init__.py
```diff
@@ -824,6 +824,22 @@ class ApplicationData(Model):
...
+    code = Unicode()
+    args = JSON()
+    log_date = DateTime()
+    # This is the identification string, is composed by a 'admin' or 'receiver_$UUIDv4'
+    subject = Unicode()
+    log_message = Unicode()
+    log_level = Int()
+    mail = Bool()
+    mail_sent = Bool()
+
```
evilaliv3: Please write here the defaults; they are useful to avoid explicit initialization if not needed and help also who read the code in understanding their usage.

evilaliv3:
@vecna 

from an fast review of (backend/globaleaks/rest/api.py) it seems to me that its missing an API for exporting the log of a precise RTip.

please remember that when we opened at this refactor putting it in the milestone it was for implementing mainly that.

Take care that with the implementation you are doing it's possible to implement https://github.com/globaleaks/GlobaLeaks/blob/openwhistleblowing/client/app/views/receiver/tiplog.html) and implement it to validate what you have done.

backend/globaleaks/db/sqlite.sql
vecna: on purpose.. mail is a boolean (when is False value 0, is the 0 considered NULL ?), and log_level can be also 0. is appropriate therefore keep the vars without constraint ?

evilaliv3: NULL is NULL (unassigned) and since now if i'm not wrong i've found that only = None = NULL

NSkelsey:
For some more info (and future me) here is what the DB looks after performing some actions with this logging subsystem.

schermata 2016-08-18 alle 4 15 33 pm

The frontend displays content in two different places.

For receivers (which is available via a little icon:
schermata 2016-08-18 alle 4 17 04 pm

For admins (settings -> Recent Activities)
schermata 2016-08-18 alle 4 17 50 pm

vecna:
That's great! @NSkelsey well done

evilaliv3:
@vodkina this pull is under review just for viwing the status in relation to the anac activities but it is not in the milestone.

evilaliv3:
@NSkelsey i lke the way you are approaching to see what is in there! well done also from my side :+1: 

what i commented above that is still to be evaluated here is basically: this implementation was done thinking mostly to debugging by the admin and not to the requirements collected in openwhistleblowing,

for this reason i've still not integrated it since ever because some refactor is needed in order to accomodate a fast execution of the main queries that will be performed by users on the system.

what would be great is to postpone this analysis to the first morning we will be together where to brainstorm collecting requirements and discuss what is already met and what is missing like the two above and then proceed with a design and integration of this work starting from mid september.

what i think i that a flat omnicomprensive table as we designed so far would not not be sufficient and too much complex to be implemented and probably not optimized for most of the queries;

saying this i'm thinking to the main different log types will require different optimized tables:
- Logs of actions on the TIP (for the recipient, custodians)
- Logs of actions of configuration by the admins
- System log of what generated by the system

in addition we would have to clearly design it around the most relevant queries,
e.g. for what i called above "Logs of actions on the TIP (for the recipient, custodians)":
-  print the log of a specific user (currently doable with a search on column 'subject' that is not in normal form 1
- print the log of a Tip (i think not doable by what i see but i may be wrong)

@NSkelsey: is it clear my point? if you want we can have a first session on this next week as i will be also

evilaliv3:
Closing as outdated.

PR #1475, opened 2015-11-09
Add client side javascript code coverage report for end to end tests.

ianribas:
Here are some changes on the build and continuous integration configuration to have code coverage of the end to end client side tests and to add this information to what is sent to [coveralls](https://coveralls.io/github/globaleaks).

With these changes it's possible to see the coverage of the client side javascript code from the end to end (protractor) tests. These results are sent to coveralls together with the coverage results of the python (trial) unit tests to have a global coverage mesure.

Change CI (travis) to run client end to end tests for coverage together with unit tests on first job and to send coverage info to coveralls aggregates in one step. For that, it is necessary to generate coveralls JSON for client coverage separately.

The coverage goes down 5% only, but the coverage of the client side is around 56%, there is just a lot more python code than client javascript.

One other thing, the workaround I had to do to have the client side sources appear in [coveralls](https://coveralls.io/github/globaleaks), makes their path a little ugly. I didn't find any way of going around this with the current tools. What could be done, of course, is generate the JSON, change all file names inside it and call the API directly. I chose not to do this yet.

Please review and tell me what you think should change.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,5 +17,9 @@ addons:
   sauce_connect: true
 install:
     - true
+before_script:
+  - "export DISPLAY=:99.0"
+  - "sh -e /etc/init.d/xvfb start"
+  - sleep 3 # give xvfb some time to start
 script:
   - $TRAVIS_BUILD_DIR/scripts/travis.sh
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,3 +1,6 @@
+Changes in version 2.60.114 - 2015-11-09
+    o [TODO]
+
 Changes in version 2.60.113 - 2015-10-22
     o Minor UI fixes
     o Packaging fixes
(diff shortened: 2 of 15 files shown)


evilaliv3:
That's great @ianribas 

Due to a revert on the relase 2.60.114 that is still not mature, i had to integrate the commit manually on devel.

Let's see how it will work and in case i will find important some refinements i will open a ticket :)

Tonight i will then refactor the Readme.md to include the results

evilaliv3:
@ianribas: i've still not tried it, but you are reporting that the total code coverage is 5% duee too a poor backend code coverage.

anyhow the current backend code coverage is 90%: https://coveralls.io/builds/4107070

is it possible that your patch is not merging results correctly?

ianribas:
Just so there are no more misunderstandings, what I meant was that the coverage ***fell*** 5%, form around 90% to around 85% when the coverage of the end to end tests were added. And, apparently, the results are being merged correctly.

Issue #1420, opened 2015-09-08
Validazione codice fiscale front-end


fpietrosanti:
Dovrà essere effettuata una validazione del codice fiscale all’atto dell’inserimento della identità nel formulario di segnalazione, tenendo in considerazione le particolari situazioni di casi di omocodia.

marceloomens:
Just a suggestion, couldn't you have a general purpose 'regex validated field'? Then you'd just need to develop a regular expression for each case with domain specific validation requirements.

evilaliv3:
@marceloomens each fied input already have the possibility to define a manual regexp :)

with this ticket we are simply thinking to implement a really specific field with specialized validation.

marceloomens:
Ah, okay, just ignore me comment then :fearful:

vodkina:
Assigned to Bruno

NSkelsey:
I came up with a solution to this problem that keeps the number of fields down. 

Instead of adding a new field type for each text based field that somebody might want to use, utilize the regular expression validation field in a different way.

<img width="367" alt="screen shot 2015-11-12 at 12 02 44 am" src="https://cloud.githubusercontent.com/assets/1095794/11110982/29780450-88d1-11e5-89fa-78f3557f6814.png">

This way if an admin needs another structured text field, we can just add an item to this list without adding to field types that everyone must wade through. The image below should show how easy this makes it to add these kinds of fields as well.

<img width="370" alt="screen shot 2015-11-12 at 12 09 46 am" src="https://cloud.githubusercontent.com/assets/1095794/11111007/b726e2da-88d1-11e5-98b1-cf909ee860ee.png">

evilaliv3:
@NSkelsey nice patch nick, really simply and effective.

any as you are identifying, relying on regexps and still giving the user visibility of the regexp even for suggested new user types is not good anyhow this would remain acceptable.
there is only one reason why it would be better to not implemten it like this: specialized fields would need to be subject to statistics and searches, and so with this proposed implementation it won't be possible to look for answers that corresponds to a SSN except from looking fo

given the work already done for having stats but also specialized fiells i would suggest in the general to go for this different road; tell me your feedback ! :)

- globaleaks has basic simple fields (inputbox, textareads, checkboxes)... https://github.com/globaleaks/GlobaLeaks/blob/master/client/app/views/admin/add_field.html#L13
- among different fields the above follow in the general basic field of type 'inputboxes'
[...]

PR #1482, opened 2015-11-12
Proposed fix for #1420

NSkelsey:
I am creating this PR to solicit feedback on a solution to #1420.
This may not achieve all of the goals of the ticket, but it addresses part of the problem.

Happy to add more changes as needed.


evilaliv3:
Closing the pull request as outdated.

Proper implementation would require integration of a specific library, while a general regexp definition possibility is already integrated inside up-to-date releases.

PR #1566, opened 2016-02-06
now it gens salt in a better and more secure way

flipchan:
i changed it so it generates salt in a more secure way using scrypt

### DIFF
--- a/backend/globaleaks/security.py
+++ b/backend/globaleaks/security.py
@@ -12,7 +12,6 @@ import base64
 import json
 import random
 import shutil
-from flask.ext.scrypt import generate_random_salt as scrypt_salt_gen #import flask version of scrypt to generate better and more secure salt
 import scrypt
 import string
 import time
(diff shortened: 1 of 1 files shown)


evilaliv3:
Thank you @flipchan for the patch but i think we should refine it a little bit before integrating it.

the patch you are suggesting is built using an additional scrypt library (globaleaks alreade uses python-scrypt) that is a plugin of Flask, that is a different framework of the one used in globaleaks that uses Twisted.

in addition why are you guessing that this would be more secure; would yo please elaborate?

flipchan:
yes ofc, atm you got def generateRandomSalt():
    """
    Return a string a-z0-9 long 32 chars (128 bit of entropy)
    """
    return ''.join(random.SystemRandom().choice(string.ascii_lowercase + string.digits) for _ in range(32)).encode('utf-8')

this is good but it will only generate low letters and numbers , scrypt will generate more different chars so scrypt would generate upper,lower,digits and special chars , yeah ur right with the flask thing , but the flask plugin for this was easy to use ^^, anyhow this should be done just using pythons own scrypt module as u said , i will c if i can solve it using the normal python-scrypt

flipchan:
check it out , i did what flask-scrypt does but with less code , should work great!

evilaliv3:
@flipchan the second patch is ok and i'm going to integrate it with some modifications.

Here is some suggestions that would important that you follow for next contributions in order to get a better integratioin:
1) if you say that something is not secure, motivate it; your code and the current one are exaly the same for a matter of security
2) remember to add the missing imports (flask was not addded to requirements.txt in first contribution), b64encode is not imported in the second so that s you can see the patch fails test; better you open branches on your repo, test them and then submitt the patch when finished.
3) your solution does not the same of what we were doing; we wanted to generate a salt of 128bit; your is even bigger but not documented and  not inline with what we have in the spec.

anyway thank you so much for this starting point and looking forward for others!

evilaliv3:
merged manually.

PR #1587, opened 2016-03-16
Feature/1076

evilaliv3:
@NSkelsey could you please take care of the review of this pull request?

All is ready for integration;
Waiting for the release of the new version of angular-zxcvbn i've applied a fix directly inside globaleaks for issue https://github.com/ghostbar/angular-zxcvbn/issues/13


client/app/views/admin/users.html
```diff
@@ -162,7 +162,8 @@
                   </div>
                   <div class="form-group">
                     <label data-translate>Password</label>
-                    <input name="password" class="form-control" data-ng-model="user.password" maxlength="{{node.maximum_namesize}}" type="password" />
+                    <input name="password" class="form-control" data-ng-model="user.password" maxlength="{{node.maximum_namesize}}" type="password" zxcvbn="passwordStrength" />
```
NSkelsey: Why keep a max password length? If you are looking for strong passwords why let the node admin force them to be weak?

client/app/index.html
```diff
@@ -195,6 +195,8 @@
     <script src="components/fusty-flow.js/src/fusty-flow.js"></script>
     <script src="components/fusty-flow.js/src/fusty-flow-factory.js"></script>
     <script src="components/ng-flow/dist/ng-flow.min.js"></script>
+    <script src="components/zxcvbn/dist/zxcvbn.js"></script>
+    <script src="components/angular-zxcvbn/dist/angular-zxcvbn.js"></script>
```
NSkelsey: @evilaliv3  neither `components/zxcvbn/dist/zxcvbn.js` or `components/angular-zxcvbn/dist/angular-zxcvbn.js` are included as files in this PR. 

In development mode running this does not work

NSkelsey:
The only drawback that I can see with integrating this is adding two external libraries. The only dependency that angular-zxvcbn has is zxvcbn which has no dependencies itself. Including the libraries as components is a good practice but I am not sure which gets used in production: the JS in components or the files downloaded from NPM.

client/app/views/admin/users.html
NSkelsey: Rerunning: 'grunt setupDepdencies'  throws: 

```
Running "bower:install" (bower) task
Fatal error: Unable to find suitable version for zxcvbn
```

evilaliv3:
When we ship the debian package, we currently build a script.js file that includes all the libraries.

Does this answer your doubt?

NSkelsey:
Rerunning: 'grunt setupDepdencies'  throws: 

```
Running "bower:install" (bower) task
Fatal error: Unable to find suitable version for zxcvbn
```

This is why Travis is dieing

client/app/views/admin/users.html
evilaliv3: 4 is the maximum rank gaved by zxcvbn.

we evaluated to accept rank 3 and 4 that are the current acceptable and strong levels

NSkelsey: I was mistaken on how the input attribute "maxlength" works in this case.

NSkelsey:
Why not move the password's do not match span into the label line.

screen shot 2016-03-16 at 5 46 07 pm

NSkelsey:
The `end2end-coverage` tests pass on my machine. I am going to merge this into devel despite the complaints from travis. I believe it is related to the git clone on line 78 of [script/build.sh](https://github.com/globaleaks/GlobaLeaks/blob/master/scripts/build.sh#L78). We shall see if merging into `devel` solves the build issue.

vodkina:
estimate, please! :)

evilaliv3:
This is a pull request yvette; sadly ZenHub does not discriminate among issues and is only the reference ticket that from my point of view deserve an estimation.

i would suggest that to every pull can stay without estimation or 0 estimation; this estimation couldto be raised if an only if the integratiton will then go longer opening discussions and  problems.

in fact its the reference ticket that already have an estimation; e.g. the 1076 already has it does not make any sens to add an estimation to this one as i also integrated it myself.

vodkina:
Well, you're right, @evilaliv3.

Issue #1589, opened 2016-03-17
Implement HTTP Basic Authentication to allow to protect the deployment during development


evilaliv3:
It has been requested by various adopters for the possibilty to protect the deployment during development in order to not disclose what under development before internal review and proper launch.

This is a feature missing since time that would be easy to be implemented by supporting HTTP Basic Authentication as by [RFC2617](https://tools.ietf.org/html/rfc2617).

The Admin should be able to enable such a feature via Advanced Settings by having the possibility for configuring a username and a password.

evilaliv3:
Implemented:
- backend implementation
- clientside UI
- migration scripts

the following are the screenshots of the UI

![screenshot from 2016-03-18 15 02 01](https://cloud.githubusercontent.com/assets/217034/13879983/81b53626-ed1a-11e5-80c0-02c940b3fe52.png)

![screenshot from 2016-03-18 15 02 01](https://cloud.githubusercontent.com/assets/217034/13880019/a89137f4-ed1a-11e5-9952-ff6f2f1857fc.png)

evilaliv3:
Here is the latest interface designed together with @NSkelsey 

we discussed that a "Danger" clarification is needed and the password should be explicilty be shown in plaintext to make it definitely clear that the feature is not implemented for security reasons.

![screenshot from 2016-03-18 19 58 52](https://cloud.githubusercontent.com/assets/217034/13888907/26b70b94-ed44-11e5-9b8a-9e615b98c11f.png)

NSkelsey:
PR #1589 closes this.

evilaliv3:
Update: this feature was abused for offering platforms to whistleblowers inside an intranet and was for this reason removed long time ago.

PR #1591, opened 2016-03-18
Feature/1589

evilaliv3:
Implementation completed!

Details on the implementation at: https://github.com/globaleaks/GlobaLeaks/issues/1589

@NSkelsey would you please review and merge this pull request?


backend/globaleaks/handlers/authentication.py
```diff
@@ -52,7 +52,7 @@ def login_whistleblower(store, receipt, using_tor2web):
     """
     login_whistleblower returns the WhistleblowerTip.id
     """
-    hashed_receipt = security.hash_password(receipt, GLSettings.memory_copy.receipt_salt)
+    hashed_receipt = security.hash_password(receipt, GLSettings.memory_copy.password_salt)
```
NSkelsey: Why not have two different salts? Instead of reusing the basic auth salt for the receipt just keep the `receipt_salt`.

evilaliv3: Cause for the receipts(all, N) we are using the same salt given the fact that they do not have a username to which assign the salt, and given that using the same salt N times or N+1 times is exactly the same :)

Any suggestion?

backend/globaleaks/db/__init__.py
```diff
@@ -54,7 +54,7 @@ def init_db(store):
 
     node = models.Node()
     node.wizard_done = GLSettings.skip_wizard
-    node.receipt_salt = generateRandomSalt()
+    node.password_salt = generateRandomSalt()
```
NSkelsey: @evilaliv3 my suggestion is to keep the `node.receipt_salt` around on this line. 

That N+1 use is different than all N other uses.

backend/globaleaks/handlers/base.py
```diff
@@ -225,6 +227,24 @@ def call_handler(cls, *args, **kwargs):
 
         return wrapper
 
+    def basic_auth(self):
```
NSkelsey: I have seen decorators in this code which make a response take a default amount of time to return. Can this be applied here?

evilaliv3: good review.

i'm going to add a commit for this.

NSkelsey:
@evilaliv3 attempting to run this PR throws an error.

`ImportError: No module named digest`

Was a file named `digest.py` left out?

backend/globaleaks/handlers/base.py
evilaliv3: discussed off the record; we agreed to not implement the protection given the intent of the feature that is for implementing a basic authentication just for limiting regular user access during development; clarifications about this are provided on the platform to users.

backend/globaleaks/db/__init__.py
evilaliv3: done!

evilaliv3:
merging it myself to speed up final integration. the review an testing will proceed on devel branch.

vodkina:
estimate, please! :)

evilaliv3:
no estimation needed as described here: https://github.com/globaleaks/GlobaLeaks/pull/1587#issuecomment-200397085

Issue #1580, opened 2016-03-11
Enable recipients to export submission from the tip list page


evilaliv3:
As an improvement for https://github.com/globaleaks/GlobaLeaks/issues/931 it would be valuable to add an export button to the tip list page.

This would make more easy for the user to decide what to download having a look to the submissions resume.

NSkelsey:
When the recipient selects multiple tips and clicks export the angular app will initiate multiple zip file downloads.

evilaliv3:
good to go

NSkelsey:
The export button can move down into the tip list itself, but the mass selection does not work. Currently this is what I have -- which is not good:

<img width="611" alt="screen shot 2016-03-18 at 7 51 35 pm" src="https://cloud.githubusercontent.com/assets/1095794/13894862/f082778a-ed42-11e5-9ff6-72e1e779537f.png">

NSkelsey:
The new format looks like this at rest:
![screen shot 2016-03-21 at 11 30 15 am](https://cloud.githubusercontent.com/assets/1095794/13923800/61a12742-ef58-11e5-80f6-3ea086e3e4b5.png)

And when hovered:

![screen shot 2016-03-21 at 11 30 30 am](https://cloud.githubusercontent.com/assets/1095794/13923808/67474672-ef58-11e5-89ea-f9e1e1f9c2cb.png)

evilaliv3:
allright :)

probably we would avoid exporting the tip links as i was doing (ret['export'] = '/rtip/' + rtip.id + '/export'
) but have the client calulate them;
this way we do not lose backend time on something that can be delegated to the client and we keep the json answers smaller?

NSkelsey:
Okay. That is fair.

NSkelsey:
What the UI looks like when delete or postpone is not enabled and the select box is hidden:

![screen shot 2016-03-21 at 11 30 22 am](https://cloud.githubusercontent.com/assets/1095794/13924809/182b17a4-ef5c-11e5-9875-1b269addf738.png)

NSkelsey:
#1595 addresses this issue.

PR #1595, opened 2016-03-21
Feature/1580

NSkelsey:
See #1580 for more discussion. 

@evilaliv3 Please review. I punted on the end2end test verifying the file download. Currently, in Firefox browsers the click brings up a 'are you sure you want to download this?' button.


evilaliv3:
Thanks @NSkelsey 

going to review the pull request;

for what relates to the Firefox issue on the end2end the issue exists quite on all browsers so that currently on saucelabs we are testing the file download only on chrome: 

for firefox there is this workaround but i've still to put it in place; i've documented it here: https://github.com/globaleaks/GlobaLeaks/issues/1596

backend/globaleaks/handlers/receiver.py
```diff
@@ -86,7 +86,7 @@ def get_receivertip_list(store, receiver_id, language):
             'preview_schema': db_get_archived_preview_schema(store, rtip.internaltip.questionnaire_hash, language),
             'preview': rtip.internaltip.preview,
             'total_score': rtip.internaltip.total_score,
-            'label': rtip.label
```
evilaliv3: i think this is a wrong merge of a comma i removed; going to remove it while mergin the pull request.

client/tests/end2end/test-globaleaks-process.js
```diff
@@ -277,6 +277,20 @@ describe('globaLeaks process', function() {
     });
   });
 
+  it('Recipient should be able to export the first submission from the tips page', function(done) {
+    login_receiver(receiver_username, receiver_password).then(function() {
+      element(by.css('#tip-0 form.tipExport button')).click().then(function() {
+        if (utils.testFileDownload()) {
```
evilaliv3: this line ```if (utils.testFileDownload()) {``` should be swapped with the previous ```element(by.css('#tip-0 form.tipExport button')).click().then(function() {``` in order to avoid the tests on the browsers that requires user input for the file download.

going to do this while merging.

client/app/js/controllers/receiver.js
```diff
@@ -7,7 +7,12 @@ GLClient.controller('ReceiverSidebarCtrl', ['$scope', '$location', function($sco
 
 GLClient.controller('ReceiverTipsCtrl', ['$scope',  '$http', '$route', '$location', '$uibModal', 'ReceiverTips',
   function($scope, $http, $route, $location, $uibModal, ReceiverTips) {
-  $scope.tips = ReceiverTips.query();
+  $scope.tips = ReceiverTips.query(function(tips) {
+      angular.forEach(tips ,function(tip) {
+          tip.export = '/rtip/' + tip.id + '/export';
+        });
+      return tips;
+  });
```
evilaliv3: i'm not certain of this; good that you tried experimenting with the promises callback but a simple ```href="{{'/rtip/' + tip.id + '/export'}}``` in the link would make it more simple and would require less overhead by the client; in fact on the ng-repeat implement automagically 1 for cycle, with your we hand in having 2 for cycle: https://github.com/globaleaks/GlobaLeaks/blob/master/client/app/views/receiver/tips.html#L176

what do you think about it?

NSkelsey: Make the change. The attribute on the form needs to be: `action="'/rtip/{{::tip.id }}/export"`

I carelessly thought that string concatenation of this form was not the style.

client/app/views/receiver/tips.html
```diff
@@ -174,13 +176,25 @@
...
+                <i class="glyphicon glyphicon-check"></i>
+              </span>
+              <span class="btn btn-xs interactive" data-ng-if="!isSelected(tip.id)" data-ng-click="tip_switch(tip.id); $event.stopPropagation();">
+                <i class="glyphicon glyphicon-unchecked"></i>
+              </span>
             </span>
+            <form class="tipExport pull-right"
+                method="POST" action="{{::tip.export }}" 
+                ng-click="console.log('blah'); alert('hello'); $event.stopPropagation(); debugger;"
+                uib-popover="{{'Export' | translate}}" popover-placement="top" popover-trigger="mouseenter"
```
evilaliv3: hello hello @NSkelsey !

client/tests/end2end/test-globaleaks-process.js
evilaliv3: the protractor selector (that inherits from webdriver) allow also nested lookups like:
```element(by.id('tip-0')).element(by.id('tip-action-export')).click().then(function() {```

Issue #1577, opened 2016-03-09
Per submission notification opt-in/opt-out (per recipient)


evilaliv3:
Various adopters reported me back that in their work they would really benefit of an opt-in/opt-out feature for the submission notification.

The typical use case of such a feature is the following:
- a submission arrive;
- the submission is checked by the recipient and the recipient not finding that much relevant decides to not delete the submission giving the whistleblower the chance to provide more data but decides to opt-out from the notifications in order to not be interrupted in its work by further discussions; the recipient will then check that submission manually logging on the platform and looking at the new material.
- the recipient would then like to be in possibility to opt-in if the submissions turns out to be relevant.

Feature request strictly related to: https://github.com/globaleaks/GlobaLeaks/issues/1576

\cc @NSkelsey @marceloomens @infosquatter @fpietrosanti

NSkelsey:
I think it is wise to do this before #1576.

Additionally, would silencing notifications mean excluding the tip from the digest? -- I think not.

evilaliv3:
I agree; this feature is more simple and would allow you to learn step by step.

In addition is a lot valuable for a user in order to not be disturbed in case of not useful communications arriving from a tip.

And i think as you that eventually the digest should include all that arrive, eventually discriminating the new-new material an the new material part of an opt out.

evilaliv3:
IE11 is now succeding;

IE10 and IE9 not; i'm going to see which are the tests that are failing now.

NSkelsey:
I created some UI to handle the display of UI notifications. The toggle button could be red, but I went with warning yellow because this is a more neutral setting.

**From the tip page this is the new resting interface:**

<img width="284" alt="screen shot 2016-03-28 at 1 05 07 pm" src="https://cloud.githubusercontent.com/assets/1095794/14083917/3126b1a2-f4e6-11e5-995e-9a7d6aaa44da.png">

**With the Silence notif button hovered:**

<img width="313" alt="screen shot 2016-03-28 at 1 05 13 pm" src="https://cloud.githubusercontent.com/assets/1095794/14083929/3c7092f8-f4e6-11e5-89c0-f3c8912f875d.png">

**With the Silence notif toggled on. (Meaning this tip will not generate notifications):**

<img width="293" alt="screen shot 2016-03-28 at 1 05 26 pm" src="https://cloud.githubusercontent.com/assets/1095794/14083941/5200069e-f4e6-11e5-8b00-b36094ccba8f.png">

[...]

PR #1614, opened 2016-03-28
Feature/1577

NSkelsey:
This addresses #1577 which gives each recipient the ability to toggle notifications for individual submissions on and off.

It uses the extensibility of the `tip.setVar( )` function on the client side and the `rtip put` handler on the server to get the job done. Finally, the actual check on whether or not to actually send the mail or not occurs in `process_mail_creation`.

I believe that `process_mail_creation` needs a bit of scrutiny. The function already does many different things and while it looked like the place to put the logic. It seems like there could be a better place to intercept the mail spooling.

### DIFF
--- a/backend/globaleaks/jobs/notification_sched.py
+++ b/backend/globaleaks/jobs/notification_sched.py
@@ -120,9 +120,15 @@ class MailGenerator(object):
         self.process_mail_creation(store, data)
 
     def process_mail_creation(self, store, data):
+        receiver_id = data['receiver']['id']
+
+        # Do not spool emails if the receiver has opted out of ntfns for this tip.
+        if not data['tip']['enable_notifications']:
+          log.debug("Discarding emails for %s due to receiver's preference." % receiver_id)
+          return
+
         # https://github.com/globaleaks/GlobaLeaks/issues/798
         # TODO: the current solution is global and configurable only by the admin
-        receiver_id = data['receiver']['id']
         sent_emails = GLSettings.get_mail_counter(receiver_id)
         if sent_emails >= GLSettings.memory_copy.notification_threshold_per_hour:
             log.debug("Discarding emails for receiver %s due to threshold already exceeded for the current hour" %
--- a/backend/globaleaks/models/__init__.py
+++ b/backend/globaleaks/models/__init__.py
@@ -315,7 +315,7 @@ class ReceiverTip(Model):
 
     new = Int(default=True)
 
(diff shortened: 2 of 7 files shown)


backend/globaleaks/models/__init__.py
```diff
@@ -315,7 +315,7 @@ class ReceiverTip(Model):
 
     new = Int(default=True)
 
-    enable_notifications = Int(default=True)
+    enable_notifications = Bool(default=True)
```
NSkelsey: @evilaliv3 please scrutinize.

evilaliv3: you are right!

client/tests/end2end/test-globaleaks-process.js
```diff
@@ -274,6 +274,23 @@ describe('globaLeaks process', function() {
     }
   });
 
+
+  fit('Recipient should be able to disable and renable email notifications', function() {
```
evilaliv3: fit() stays for fitness?

client/tests/end2end/test-globaleaks-process.js
```diff
@@ -274,6 +274,23 @@ describe('globaLeaks process', function() {
...
+    login_receiver(receiver_username, receiver_password);
+    element(by.id('tip-0')).click();
+    var silence = element(by.id('tip-action-silence'));
+    silence.click();
+    var notif = element(by.id('tip-action-notify'));
+    notif.evaluate('tip.enable_notifications').then(function(enabled) {
+      expect(enabled).toEqual(false);
```
evilaliv3: i like this fine checks!

backend/globaleaks/jobs/notification_sched.py
```diff
@@ -120,9 +120,15 @@ def process_ReceiverFile(self, store, rfile, data):
         self.process_mail_creation(store, data)
 
     def process_mail_creation(self, store, data):
+        receiver_id = data['receiver']['id']
+
+        # Do not spool emails if the receiver has opted out of ntfns for this tip.
+        if not data['tip']['enable_notifications']:
+          log.debug("Discarding emails for %s due to receiver's preference." % receiver_id)
```
evilaliv3: 'receiver' is generally nowadays used only in the code; if the message is for the user(of the webapplication or the admin that looks the logs) the texts should prefer using 'recipient'. various translators  indipendently suggested this.

evilaliv3:
Well done @NSkelsey !

PR #1616, opened 2016-03-30
Feature/1565

NSkelsey:
This is not quite done. The feature still needs test coverage and error handling, but I am creating the PR now to pick up comments before I look at it tomorrow.

The one bit that I am unsure of how to do after a fair amount of sleuthing is generating the success and failure popups in the frontend (the one that comes up when you are not properly authenticated). If the post returns a non 200 status code, an error popup with a bit of debug info should display.


backend/globaleaks/handlers/admin/notification.py
```diff
@@ -157,3 +159,33 @@ def put(self):
...
+                               GLSettings.memory_copy.default_language)
+      send_to = user['mail_address']
+
+      try: 
+        log.debug("Attempting to send test email to:"+ send_to)
+        msg = "This is a test email generated by the GlobaLeaks administration interface.\n"+\
+        "If you were not expecting this email, please contact the owner of this email address."
+        result = yield sendmail(send_to, 'Test Email Config', msg)  
+        self.set_status(200)
+      except Exception as e:
```
NSkelsey: @evilaliv3 this is not the way you usually handle exceptions. Could I get some guidance here?

evilaliv3:
Nice pull request @NSkelsey;

Trying to answer your doubts i think that in order to proceed we need:
1) remember to keep it simple!
2) the test button should send an email on the user triggering the test email; i would keep it synchronous with a specific API as you have than that would return success or failure (there is no issue in having a long waiting API as the functionality is simply to keep the connection open, and this would avoid us to save the status of the smtp server on the db along with the email of the recipient and other infos; the feature is only for validating the email configuration, not mail schedulers or other stuffs.
3) i think you are missing to design the test email;  and for this we need a title and a text to be put under internationalization:
something like:
Title: Test email for the email notification settings
Text: This email is ...... 

For those templates i would suggest to not let the user customize them through the platform so they can be simple without using any variable as other templates do.

backend/globaleaks/handlers/admin/notification.py
```diff
@@ -157,3 +159,33 @@ def put(self):
...
+    def post(self):
+      """
+      Errors: InvalidAuthentication, ForbiddenOperation, TODOstmpError
+      """
+      user = yield get_user_settings(self.current_user.user_id, 
+                               GLSettings.memory_copy.default_language)
+      send_to = user['mail_address']
+
+      try: 
+        log.debug("Attempting to send test email to:"+ send_to)
```
evilaliv3: missing space here between " and +

backend/globaleaks/handlers/admin/notification.py
```diff
@@ -157,3 +159,33 @@ def put(self):
...
+
+    @BaseHandler.transport_security_check('admin')
+    @BaseHandler.authenticated('admin')
+    @inlineCallbacks
+    def post(self):
+      """
+      Errors: InvalidAuthentication, ForbiddenOperation, TODOstmpError
+      """
+      user = yield get_user_settings(self.current_user.user_id, 
+                               GLSettings.memory_copy.default_language)
```
evilaliv3: The alignment we have used so far is with the [G]Lsettigs aligned with [s]elf.current but i do not know if it's standards or not

backend/globaleaks/handlers/admin/notification.py
```diff
@@ -157,3 +159,33 @@ def put(self):
...
+
+      try: 
+        log.debug("Attempting to send test email to:"+ send_to)
+        msg = "This is a test email generated by the GlobaLeaks administration interface.\n"+\
+        "If you were not expecting this email, please contact the owner of this email address."
+        result = yield sendmail(send_to, 'Test Email Config', msg)  
+        self.set_status(200)
+      except Exception as e:
+        log.debug("Sending test email failed.")
+        self.set_status(400) # 402 Unprocessable entity threw a strange error.
```
evilaliv3: answering your questions in relation to the exceptions, currently all SMTP exceptions hare handled internally without returning notihing well organized in "result = yield sendmail" so that i think that currently no exceptions will reach the ```except Exception as e``` clause.

i think that what we would need to do is to decide the format of ```result``` for internal communications in general.

we could use a format like:
```
{
  'status': status_value'
  'message': 'eventual_message'
}
```

so that instead of a try catch we will have to analyze the status and if !ok print something.

backend/globaleaks/handlers/admin/notification.py
NSkelsey: It is standard according to [PEP8](https://www.python.org/dev/peps/pep-0008/#indentation)

NSkelsey: It is standard according to [PEP8](https://www.python.org/dev/peps/pep-0008/#indentation).

NSkelsey:
1. Yep, will do.

2. Requests to /admin/notification/mail will block until the db and send_mail respond.

3. Okay, I can take what is there and hook internationalization in. This is what the emails look like in english:

>
screen shot 2016-03-30 at 10 09 01 am

backend/globaleaks/handlers/admin/notification.py
evilaliv3: perfect!

evilaliv3:
Perfect! as the email apply to GlobaLeaks whatever it will be rebranded and in situatiions where the adopter does not want to see the terrible 'GlobaLeaks' name always refer to "the platform". the email texts should be changed accordingly! ;)

NSkelsey:
@evilaliv3 I have run into two blocking issues here that I cannot resolve. 

1) Running `grunt makeAppData` yields:
```bash
> grunt makeAppData
Running "makeAppData" task
Warning: ENOENT: no such file or directory, open 'pot/en.po' Use --force to continue.

Aborted due to warnings.
```
This seems like an uncommitted file.

2) The notif object returned by `yield get_notification(user['language'])` returns empty strings for any new localized field I try to add. Trying to forcefully insert valid JSON to change the localized fields using `sqlite3 workingdir/db/glbackend-31.db` did not work either. 

As a result of that problem (and to stick with the naming conventions @_@) I renamed the `test_email_static_title` variable to the verbose yet ambiguous `admin_test_static_mail_title` 

Not being able to properly set the localized values in the DB may go away if I can build the translations locally. I may be missing some obvious piece as well, but the commits along with the thinking in this branch has turned into spaghetti.

evilaliv3:
you are right nick; currently that file is generated as one runs ```grun updateTranslationsSource```

BEAWARE: do not run it now as running it can cause to destroy translations on transifex.

i'm going to split updateTranslationsSource with two different commands:
- grunt 'createTranslationSource' <- that one will use for the testing
- grunt 'pushTranslationSource' <- that one will use in order to put texts under internationalization on transifex

evilaliv3:
doing it now and then i will rebase this pull on the current branch with the fixes you need to proceed.

evilaliv3:
@NSkelsey: this commits implement what you were requiring:
https://github.com/globaleaks/GlobaLeaks/commit/f41bb2210af4a92a1b741ff8e969671e6762b298

you can now simply run:
- ```grunt makeTranslationsSource```: to create your pot/en.po
- ```grunt updateTranslations```: to create the application data fetching the new translation from transifex

you can do that with your transifex account by having a .transifexrc:
```
[https://www.transifex.com]
hostname = https://www.transifex.com
username = 
password = 
```

remember to NOT input ```grunt pushTranslationsSource"; this is the command that update the pot/en.po on transifex and should be executed only on the /devel branch; anyhow if you do it you will get a confirmatiion dialog.

NSkelsey:
@evilaliv3 okay, those commands seem to work. 

The next step is creating `'/Users/nskelsey/.transifexrc'`.

client/app/views/admin/mail.html
```diff
@@ -17,6 +18,14 @@
               <i class="glyphicon glyphicon-ok"></i>
               <span data-translate>Save</span>
             </button>
+            <input type="hidden" name="session" value="{{session.id}}" />
+            <button uib-popover="{{'Send a test email from the SMTP account to your email address.' | translate}}"
+                    popover-placement="right"
+                    popover-trigger="mouseenter"
+                    ng-click="updateThenTestMail();"
+                    class="btn btn-success">
+              Test Email <i class="glyphicon glyphicon-export"></i>
```
NSkelsey: Needs Translate directive.

PR #1617, opened 2016-03-30
Remove redundant usages of request.set_status(200) and request.finish()

evilaliv3:
This pull requests apply the following changes to all the handlers of the application:

- remove redundant self.finish() usages.
- replace self.finish(payload) with self.write(payload)
- remove redundant self.set_status(200)

@NSkelsey would you please check if this is ok for you?

### DIFF
--- a/backend/globaleaks/handlers/admin/langfiles.py
+++ b/backend/globaleaks/handlers/admin/langfiles.py
@@ -37,7 +37,6 @@ class AdminLanguageFileHandler(BaseHandler):
     @BaseHandler.authenticated('admin')
     def get(self, lang):
         self.set_status(204)
-        self.finish()
 
     @BaseHandler.transport_security_check('admin')
     @BaseHandler.authenticated('admin')
@@ -49,7 +48,6 @@ class AdminLanguageFileHandler(BaseHandler):
         uploaded_file = self.get_file_upload()
         if uploaded_file is None:
             self.set_status(201)
-            self.finish()
             return
 
         path = self.custom_langfile_path(lang)
@@ -67,7 +65,6 @@ class AdminLanguageFileHandler(BaseHandler):
         log.debug("Admin uploaded new lang file: %s" % dumped_file['filename'])
 
         self.set_status(201)  # Created
-        self.finish()
 
     @BaseHandler.transport_security_check('admin')
     @BaseHandler.authenticated('admin')
@@ -83,6 +80,3 @@ class AdminLanguageFileHandler(BaseHandler):
             raise errors.LangFileNotFound
 
         os.remove(path)
-
-        self.set_status(200)
-        self.finish()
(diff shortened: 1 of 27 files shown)


backend/globaleaks/handlers/admin/langfiles.py
```diff
@@ -49,7 +48,6 @@ def post(self, lang):
         uploaded_file = self.get_file_upload()
         if uploaded_file is None:
             self.set_status(201)
-            self.finish()
             return
```
NSkelsey: Shouldn't this throw an error? The file is none, nothing was created.

backend/globaleaks/handlers/admin/staticfiles.py
```diff
@@ -121,7 +121,6 @@ def post(self, filename):
         uploaded_file = self.get_file_upload()
         if uploaded_file is None:
             self.set_status(201)
-            self.finish()
             return
```
NSkelsey: Similar to the other return above. This return closes execution, but the status code should be a 40x not a 201.

Additionally, if this pattern occurs a few times why not make it a function.

backend/globaleaks/handlers/rtip.py
```diff
@@ -399,7 +398,6 @@ def put(self, tip_id):
 
         # TODO A 202 is returned regardless of whether or not an update was performed.
         self.set_status(202)  # Updated
```
NSkelsey: It would be nice to see an example of the proper handling of failure codes in a handler like this one.

NSkelsey:
Looks good to me. I would like to see one function written that handles static file uploads, and a good example of custom error codes being used well, but other than that ACK to merge.

backend/globaleaks/handlers/admin/langfiles.py
evilaliv3: Each of these requests is for uploading a chunk of the file and flow expects 200, 201 or 202 to say that the single chunk has been loaded correctly.

from https://github.com/flowjs/flow.js:

For every request, you can confirm reception in HTTP status codes (can be change through the permanentErrors option):

200, 201, 202: The chunk was accepted and correct. No need to re-upload.
404, 415. 500, 501: The file for which the chunk was uploaded is not supported, cancel the entire upload.
Anything else: Something went wrong, but try reuploading the file.

backend/globaleaks/handlers/admin/staticfiles.py
evilaliv3: see above comment; i'ts a flow spec. is not an error condition but simply a condition where the fille is still not completed;

flow.js specs wants 200, 201 or 202

backend/globaleaks/handlers/rtip.py
evilaliv3: what do you mean with proper handling of failure codes? could you clarify?

NSkelsey: So, in this function `RTipInstance.put` reachable via `/receiver/{{ tip_id }}` the PUT request could contain invalid options and fail within the put handler.

However, as it is written, if the operation equals something other than "set" or "postpone" the handler still sets the response to 202. Clearly, the PUT operation did nothing. Shouldn't there be an `else: self.set_status(40x)` clause that catches these cases.

Issue #1606, opened 2016-03-25
The text that prompt to login whistleblowers should be customizable


evilaliv3:
An adopter requested for the possibility that the text that prompt to login whistleblowers could be customizable.

Default text will continue for now to be: "Have you already filed a submission? Enter your key code."

As discussed with @NSkelsey probably the title for this admin configuration should be: "Text of prompt to login whistleblowers".

This text has been already pushed for translations.

evilaliv3:
Implemented and integrated in devel; will be shipped in next release.

PR #1618, opened 2016-03-30
Implement feature #1606

evilaliv3:
@NSkelsey: this is an other full example of feature that required:
- addition of new variable to the db
- import/export of that variable via a REST API
- implementation of a simple UI
- implementation of the migration script

would you please review and merge this?


backend/globaleaks/db/migrations/update_31/__init__.py
```diff
@@ -182,6 +189,9 @@ def migrate_Node(self):
                 new_node.contexts_clarification = old_node.context_selector_label
                 continue
 
+            if v.name == 'whistleblowing_receipt_prompt':
+                continue
```
NSkelsey: @evilaliv3 Is the default value for `node.whistleblowing_receipt_prompt` set in`self.update_model_with_new_templates`?

If not you need to assign it. Otherwise I will merge!

evilaliv3: it is in the appdata dictionary that is freshly loaded from the appdada.json file!

evilaliv3:
i added the milestone for error; pull requests if not demanding that much integration work does not require to be putted into milestone if they have already an associated ticket.

PR #1627, opened 2016-04-08
Feature/1626

NSkelsey:
This PR introduces basic support for the angular-file-saver `FileSaver.saveAs` function.

It adds three new scripts to be included in frontend JS:

```html
<script src="components/file-saver.js/FileSaver.js"></script>
<script src="components/blob-polyfill/Blob.js"></script>
<script src="components/angular-file-saver/dist/angular-file-saver.min.js"></script>
```

### DIFF
--- a/backend/globaleaks/handlers/export.py
+++ b/backend/globaleaks/handlers/export.py
@@ -71,19 +71,6 @@ class ExportHandler(BaseHandler):
     @BaseHandler.transport_security_check('receiver')
     @BaseHandler.authenticated('receiver')
     @inlineCallbacks
-    def post(self, rtip_id):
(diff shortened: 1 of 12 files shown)


NSkelsey:
There is a known issue on `FileSaver.saveAs()` with `Safari 8.02`. In the error console, this is what we get when clicking the button. (nothing downloads)

screen shot 2016-04-08 at 3 11 19 pm

evilaliv3:
could you link to the issue?

being the only valid library, in the meantime possibly they will fix it. please proceed considering other working browsers that could allow you anyhow to accomplish the workflow; we will then understand how to handle this in the short term.

PR #1646, opened 2016-04-18
Feature/pgp key validation

NSkelsey:
For this branch to merge into `devel`:

- The tests that explicitly set the recipient's pgp key in the UI must pass.
- The key_info parser in the backend must be removed.
- The DB schema must be updated to reflect the deletion of the key_info field.


NSkelsey:
Wrong branch.... damn!

PR #1647, opened 2016-04-18
Feature/pgp key validation

NSkelsey:
For this branch to merge into `devel`:

- The tests that explicitly set the recipient's pgp key in the UI must pass.
- The key_info parser in the backend must be removed.
- The DB schema must be updated to reflect the deletion of the key_info field.
- Load the parsed key into memory via a `GLBrowserCrypto` service

### Backend Code Reduction

`Parse_pgp_key_info` in users.py must be cleaned out.

Files that use this info are:
- [handlers/admin/user.py](https://github.com/globaleaks/GlobaLeaks/blob/master/backend/globaleaks/handlers/user.py)
- [handlers/admin/receiver.py](https://github.com/globaleaks/GlobaLeaks/blob/master/backend/globaleaks/handlers/admin/receiver.py)


evilaliv3:
Well done @NSkelsey;

i just did a commit of review, could you please see it and complete what you already annotated plus exporting the PGP key and using it at runtime for showing the key infos?

in a minute i will annotate the various changes i did so that you can review them and tell me what you think about the changes :)

client/app/js/crypto/module.js
```diff
@@ -119,20 +122,26 @@ angular.module('GLBrowserCrypto', [])
...
   function validatePubKey(textInput) {
     // Check for obvious problems.
     if (typeof textInput !== 'string') {
       return false; 
     }
+
     var s = textInput.trim();
+
+    if (s === '') {
+      return true;
```
NSkelsey: This is return should be false. The pub-key is not valid

client/app/js/crypto/module.js
```diff
@@ -63,7 +67,7 @@ angular.module('GLBrowserCrypto', [])
   }
 
   return {
-    restrict: 'E',
+    restrict: 'A',
```
evilaliv3: generally is preferrable to use standard html elements (e.g. <div>) and try to design our elements so that are compatible with html (e.g. lets see for our custom if we can go for the data-* that pass the validators/wai-aria stuff)

client/app/js/crypto/module.js
```diff
@@ -83,28 +87,27 @@ angular.module('GLBrowserCrypto', [])
...
+      required: '&attachedRequired',
     },
     controller: ['$scope', function($scope) {
-      
-      // The key point of the controller occurs when the keyForm._pubKeyInp input 
+      // The key point of the controller occurs when the keyForm.pgpkey input
       // becomes valid. When that happens the public key attached via localModel
       // is now both ready for use elsewhere in the application and for display
-      $scope.$watch('keyForm._pubKeyInp.$valid', function(newVal, oldVal) {
+      $scope.$watch('keyForm.pgpkey.$valid', function(newVal, oldVal) {
```
evilaliv3: i know that you liked to keep this private but i think that anyhow we have to make possible for every for for investigating this inside.

client/app/js/crypto/module.js
```diff
@@ -83,28 +87,27 @@ angular.module('GLBrowserCrypto', [])
...
 .directive('pgpPubkeyInput', function() {
-
   return {
-    restrict: 'E',
+    restrict: 'A',
     templateUrl: '/views/partials/pgp/pubkey_input.html',
     scope: {
       localModel: '=attachedModel',
       keyForm: '=attachedForm',
+      required: '&attachedRequired',
```
evilaliv3: this should be a direct binding; let's see if it work correctly

client/app/js/crypto/module.js
evilaliv3: yep i totally know.

the reason i did the following is that using this wizard allow empty keys and marking the form invalid caused pain :)

but i now got an idea that if solves the issue will make me remove this immediatly.

going to comment the code below and then pushing a small commit.

client/app/views/partials/pgp/pubkey_display.html
```diff
@@ -1,9 +1,9 @@
 <label data-translate>Details of the PGP key:</label>
 <p>
-  <span data-translate>Fingerprint: </span>
+  <span data-translate>Fingerprint:</span>
   <b>
     {{ key_details.fingerprint.slice(0,25) }}
-    &nbsp
+    &nbsp;
```
evilaliv3: ; !

client/app/views/partials/pgp/pubkey_input.html
```diff
@@ -1,25 +1,18 @@
...
+  <textarea
+    pgp-pubkey-validator
+    class="form-control"
+    name="pgpkey"
+    data-ng-attr-placeholder="{{'Copy and paste the PGP public key here'}}"
+    data-ng-model="localModel"
+    data-ng-required="required"></textarea>
+  <br />
+  <div class="alert alert-danger"
+       data-ng-show="keyForm.pgpkey.$error.pgpPubKeyValidator && keyForm.pgpkey.$modelValue.$invalid">
```
evilaliv3: better to use $invalid than $dirty.

if i write "a" then i remove "a" that remain marked dirty anyhow.

client/app/views/partials/preferences/pgpkey.html
```diff
@@ -1,30 +1,31 @@
...
+      <label data-translate>Remove the PGP key</label>
+    </div>
   </div>
-</div>
 
-<div class="form-group" data-ng-hide="preferences.pgp_key_status === 'enabled'">
-  <div class="alert alert-danger">
-    <div data-translate>Encryption has not been enabled!</div>
-    <div data-translate>Email notifications won't be encrypted.</div>
+  <div class="form-group" data-ng-hide="preferences.pgp_key_status === 'enabled'" data-ng-class="{'has-error': !pgpkeyForm.$valid}">
```
evilaliv3: added the classes for showing the red borders in of missing required / invalid key.

client/app/views/partials/preferences/pgpkey.html
```diff
@@ -1,30 +1,31 @@
...
+    <div>
+      <label data-translate>Set a PGP public key for encryption of email notifications</label>
+      <div pgp-pubkey-input attached-model="preferences.pgp_key_public" attached-form="pgpkeyForm"></div>
+    </div>
   </div>
-  <div>
-    <label data-translate>Set a PGP public key for encryption of email notifications</label>
-    <pgp-pubkey-input attached-model="preferences.pgp_key_public" attached-form="preferencesForm" ></pgp-pubkey-input>
+  <div class="form-group">
+    <button type="submit" class="btn btn-success" data-ng-click="pref_save()" data-ng-disabled="pgpkeyForm.$invalid">
```
evilaliv3: disabled the button when an invalid key is present

client/app/views/wizard/main.html
```diff
@@ -154,7 +156,7 @@
       </div>
       <div class="col-md-6">
         <p class="wizard-help-hint">
-          <span data-translate>In order to deliver encrypted emails and files each recipient needs an associated PGP key. Cut and paste the key of the recipient or mark the checkbox:</span>
+          <span data-translate>In order to deliver encrypted emails and files each recipient needs an associated PGP key.</span>
```
evilaliv3: simplified the texts that was too messy; but this is an other story not related to your lovely work! ;)

client/app/views/wizard/main.html
```diff
@@ -135,14 +135,16 @@
...
-            <textarea name="pgpkey" class="form-control" data-ng-attr-placeholder="{{'Copy and paste the PGP public key here' | translate}}" data-ng-model="receiver.pgp_key_public" data-ng-required="!admin.node.allow_unencrypted"></textarea>
-            <input data-ng-model="admin.node.allow_unencrypted" data-ng-click="open_modal_allow_unencrypted()" type="checkbox" data-ng-change="receiver.pgp_key_public = admin.node.allow_unencrypted ? '' : receiver.pgp_key_public">
-            <span data-translate>Enable insecure configuration without PGP encryption.</span>
-            <div data-ng-show="admin.node.allow_unencrypted">
-              <span class="label label-danger" data-translate>Danger</span>
-              <span data-translate>By enabling this option, the system will be storing the submitted files unencrypted on your server.</span>
-              <span data-translate>This happens whenever encryption is unavailable.</span>
+          <div class="col-md-8" data-ng-class="{'has-error': wizardForm.pgpkey.$error.required || (receiver.pgp_key_public !== '' && wizardForm.pgpkey.$invalid)}">{{wizardForm.pgpkey.$invalid}}
```
evilaliv3: ops i forgotten a debug output here; going to remove this now

client/app/js/crypto/module.js
NSkelsey: I thought a bit about this and I am cool with this solution. There is definitely a general solution in this case, I think there is a good way to handle arbitrary validators attached to a directive like this. I just don't quite know what it is.

client/app/views/partials/pgp/pubkey_display.html
```diff
@@ -1,9 +1,9 @@
 <label data-translate>Details of the PGP key:</label>
 <p>
-  <span data-translate>Fingerprint: </span>
+  <span data-translate>Fingerprint:</span>
```
evilaliv3: no spaces at the start or end of the internationalized strings; eventually lets keep them outside

is really difficult to that translators whill leave them

client/app/js/crypto/module.js
NSkelsey: ACK, agreed.

PR #1649, opened 2016-04-19
Feature/pgp key validation

NSkelsey:
Created new pull to clean out the old log.

For this branch to merge into devel:

- ~~The tests that explicitly set the recipient's pgp key in the UI must pass.~~
- ~~The key_info parser in the backend must be removed.~~
- ~~The DB schema must be updated to reflect the deletion of the key_info field.~~
- The DB must be migrated to disregard the newly deprecated fields
- Load the parsed key into memory via a GLBrowserCrypto service


client/app/views/wizard/main.html
```diff
@@ -109,7 +109,7 @@
           </label>
           <div class="col-md-8" data-ng-class="{'has-error': wizardForm.receiver_email.$error.required || wizardForm.admin_email.$error.pattern}">
             <input name="receiver_email" class="form-control" data-ng-model="receiver.mail_address"  maxlength="{{node.maximum_namesize}}" data-ng-pattern="email_regexp" type="text" data-ng-attr-placeholder="{{'Enter the email address used to notify the recipient' | translate}}" required>
-            <div class="text-danger" data-ng-show="wizardForm.receiver_email.$error.pattern">
+            <div class="text-red" data-ng-show="wizardForm.receiver_email.$error.pattern">
```
evilaliv3: regression here but i will check this at the time of the merge

backend/globaleaks/tests/handlers/test_notification.py
```diff
@@ -18,8 +18,11 @@ class TestNotificationInstance(helpers.TestHandlerWithPopulatedDB):
     @inlineCallbacks
     def test_get(self):
         handler = self.request(role='admin')
+
         yield handler.get()
+
         self.assertEqual(self.responses[0]['server'], 'demo.globaleaks.org')
+        self.assertTrue('exception_email_pgp_key_info' not in self.responses[0])
```
evilaliv3: i'm more for implementing a retesting that based on data definitions in requests.py will check presence of all keys expected to be in the REST interface and would check that no additional keys would be there.

we can have a ```self.ValidateDataStructure(dictionary, dictionaryDef)``` were to use a dictionaryDef from request.py and eventually arrive to have soomething aotomagic from models.py where to have filters depending on users role but this part only in the future future! :)

evilaliv3:
well done @NSkelsey !

the migration script was working perfectly!

for now feel free to merge it if you are confident with my changes.

i've updated the test_user.py test by using the validation we were using so far;
instead of adding the "check key not in dict" check you added i would like that in a bit we could trasform this "delete key from dict" into a "raise ExceptionSomething":
https://github.com/globaleaks/GlobaLeaks/blob/master/backend/globaleaks/handlers/base.py#L382

backend/globaleaks/tests/handlers/test_notification.py
```diff
@@ -22,7 +24,7 @@ def test_get(self):
         yield handler.get()
 
         self.assertEqual(self.responses[0]['server'], 'demo.globaleaks.org')
-        self.assertTrue('exception_email_pgp_key_info' not in self.responses[0])
+        self._handler.validate_message(json.dumps(self.responses[0]), requests.AdminNotificationDesc)
```
NSkelsey: Nice. I assume the validate_message throws on failure.

NSkelsey:
Review looks good. Merging.

Issue #1641, opened 2016-04-14
The DatePicker UI should be subjected to internationalization


evilaliv3:
The DatePicker UI is currently not subject to internationalization.

This ticket is to keep track of the research on the feasibility of this and the possible implementation.

The requirement has been requested in the context of the ICC pilot.

evilaliv3:
This should be probably feasible by using simply [```angular-i18n```](https://github.com/angular/bower-angular-i18n) that is the official module of Angular for the internatonalization.

evilaliv3:
Angular does now support naturally to swap the locale at runtime so that we should evaluate using: ```https://github.com/lgalfaso/angular-dynamic-locale``` by @lgalfaso

We have to see how to make it possible inlining the locales in the built script and this is tracked by: https://github.com/lgalfaso/angular-dynamic-locale/issues/93

NSkelsey:
1) angular has an i18n support: great!
2) this support is static: oh dear!
3) there is a good simple module making it dynamic, https://github.com/lgalfaso/angular-dynamic-locale: GREAT!
4) this module loads a separate file at runtime each time you change language: OH DEAR!
5) ok doing some inlining we got out of this mess!!!! finally !!!! yeah!!!! ou yeeeeaaa!
there is something we have to think about translations then in general

NSkelsey:
@evilaliv3 I can confirm that using `angular-dynamic-locale` works for changing the UI within datepickers and datepicker-popups.

I will push a functioning sandboxed example as a separate branch next to devel. The question for me, is how to integrate this into the inlining....!

NSkelsey:
@evilaliv3 branch `feature/dynamic-locale` has a functioning example of setting the $locale object to french after the fact. I do not know how the you intend to implement the support for the ICC's request so I am leaving this here.

These are the things left to clarify:
1. At what point must the locale be set for a specific user? At first visit or after they choose a language from a list?
2. How can we package the angular-i18n/en-us.js files into the existing js bundle?

evilaliv3:
@NSkelsey most of the analysis on this is at ticket https://github.com/globaleaks/GlobaLeaks/issues/1445 that would be nice if you could address together with this :)

#1445 talks about the wizard but it apply in the real to every user navigation session.

this would help you aswering your point 1. and the order of preference for me continue to be:
1) explicit url lang (?lange=en),
2) receiver/admin language if explicitly set by the user (this as far far improvement as it require to think a lot about this, i think)
[...]

Issue #1192, opened 2015-03-13
Languages selector should be internationalized


evilaliv3:
It has been suggested by translators and UI experts that also languages names should be internationalized so that depending on the user selected language also the language selector is understandable by the user.

e.g. for me that i'm italian, the "English" selector should appear "Inglese".

This would require extracting languages names and pushing them on transifex.

fpietrosanti:
I'd suggest to evaluate if there is some library/archive that already do provide that feature (ie: name of languages in most of world languages), without a translator having to work over it.

evilaliv3:
it was already evaluated by me and that does not exist.

few libraries exists and not complete and in addition our translations are tightly mapped ont transifex names, and it's totally impossible that some library uses exactly the same names as starting point.

i will interact with the transifex developers in order to offer such a feature in general and share this translators effort on all the transifex projects.

fpietrosanti:
Consider also asking this to OTF's mailing list, as a shared deliverable that many OTF projects will benefit from

eighthave:
I think the best way to handle the names of the languages is to keep in their native language only, and never translate them into any language (including English).

To take your Italian example, if you are reading the Italian text and you don't understand what Ελληνικά means, then that language option is not going to be useful for you.  But if you only speak Ελληνικά, then it is going to be a lot easier to read the word "Ελληνικά" than "greek" or "greco".  Just like if you were reading the greek text, you'll find it a lot easier to read "Italiano" than "ιταλικά".

For example, here is what we post in the app description for ChatSecure:

★ WE SPEAK YOUR LANGUAGE: ChatSecure is available for friends who speak: العربية, Български, 中文(简体), 中文(台灣), Català, Čeština, Dansk, ‏עברית‏, Nederlands, English, Esperanto, فارسی, Français, Deutsch, Ελληνικά, Magyar, Bahasa Indonesia, Italiano, 日本語, 조선말, Norsk, Polski, Português, Русский, español, español (España), Slovenčina, Slovenščina, Српски, Svensk, བོད་སྐད།, Türkçe, and Tiếng Việt

eighthave:
As for getting the names of languages in the native language, you can get them directly from Java/Android:

https://github.com/guardianproject/ChatSecureAndroid/blob/master/src/info/guardianproject/util/Languages.java

evilaliv3:
thank you @eighthave 
your suggestion is gret.

[...]

PR #1657, opened 2016-04-28
Feature/dynamic locale

NSkelsey:
Creating a Pull Request just to collect comments on the whole diff, not specific commits.

Also proof that #1641 is solved by this:

screen shot 2016-04-27 at 10 57 25 pm

This pull needs to address #1192 as well.


client/app/js/controllers/main.js
```diff
@@ -549,7 +524,7 @@ controller('DisableEncryptionCtrl', ['$scope', '$uibModalInstance', function($sc
       $uibModalInstance.close(true);
     };
 }]).
-controller('IntroCtrl', ['$scope', '$rootScope', '$uibModalInstance', function ($scope, $rootScope, $uibModalInstance) {
+controller('IntroCtrl', ['$scope', '$rootScope', '$uibModalInstance', 'GLTranslate', function ($scope, $rootScope, $uibModalInstance, GLTranslate) {
```
NSkelsey: @evilaliv3 I believe the IntroCtrl is not used anywhere in the code. Can we delete it?

evilaliv3: Yep, let's remove it as it is something used only in the amnesty project and never that integrated.

as you had to touch it in this branch and it is a removal ok for remove this here in this branch.

client/app/js/controllers/main.js
```diff
@@ -1,21 +1,17 @@
...
+GLClient.controller('MainCtrl', ['$q', '$scope', '$rootScope', '$http', '$route', '$routeParams', '$location',  '$filter', '$translate', '$uibModal', '$timeout', 'Authentication', 'Node', 'Contexts', 'Receivers', 'WhistleblowerTip', 'fieldUtilities', 'GLCache', 'GLTranslate',
+  function($q, $scope, $rootScope, $http, $route, $routeParams, $location, $filter, $translate, $uibModal, $timeout, Authentication, Node, Contexts, Receivers, WhistleblowerTip, fieldUtilities, GLCache, GLTranslate) {
     $rootScope.started = false;
     $rootScope.showLoadingPanel = false;
     $rootScope.successes = [];
     $rootScope.errors = [];
 
-    $scope.rtl = false;
-
-    $rootScope.language = $location.search().lang;
```
evilaliv3: well done to move things to service and in general avoid the abuse of rootScope when not needed. well done.

client/app/js/controllers/wizard.js
```diff
@@ -1,7 +1,7 @@
 GLClient.controller('WizardCtrl', ['$scope', '$rootScope', '$location', '$route', '$http', '$uibModal', 'Admin',
-                    'DefaultAppdata', 'CONSTANTS',
+                    'DefaultAppdata', 'GLTranslate', 'CONSTANTS',
```
evilaliv3: instead of passing it to all controllers that would need it probably we can inject it in the MainCtrl only, this way it will be inherited in the scope of all the other.

client/app/js/services.js
```diff
@@ -914,4 +918,121 @@ angular.module('GLServices', ['ngResource']).
...
+    AddDefaultLang: function(lang) {
+      facts.nodeDefault = lang;
+      determineLanguage();
+    },
+
+    AddUserPreference: function(lang) {
+      facts.userPreference = lang;
+      determineLanguage();
+    },
+
```
evilaliv3: avoid trailin space line

client/app/js/controllers/main.js
```diff
@@ -576,15 +551,7 @@ controller('IntroCtrl', ['$scope', '$rootScope', '$uibModalInstance', function (
...
-  $scope.data = {
-    'language': $scope.language
-  };
-
-  $scope.$watch("data.language", function (newVal, oldVal) {
-    if (newVal && newVal !== oldVal) {
-      $rootScope.language = $scope.data.language;
-    }
-  });
+  $scope.GLTranslate = GLTranslate;
```
evilaliv3: probably this is the only line needed that would make every child scope inherit access to the object.

client/app/index.html
```diff
@@ -6,7 +6,6 @@
     <meta name="description" content="" />
     <meta name="viewport" content="width=device-width, user-scalable=no" />
     <meta name="referrer" content="never" />
-    <base href="/" />
```
evilaliv3: yep for now @NSkelsey do not mind at this; i also got various issues that i now do not remember but let's keep that hashbang for now :)

evilaliv3:
nice work @NSkelsey;

ok if you proceed on the languages mapping (for angular is easy as we know the code we use (the transifex one) and the language offered by angular-i18n;
For the detection of the language i think you should proceed testing what you get with (https://developer.mozilla.org/de/docs/Web/API/NavigatorLanguage/language) that maybe would require for you to evaluate an other mapping

transifex has: zh_CN
we can expect strange stuff like Zh-Cn and it should work

i would suggest to implement a basic manipulation like:
```
1) always replace - in _ not bothering if it present or not.

2) if - is present lowercase the first part, UPPERCASE the second part
  2a) check if the whole text is in the enabled languages
  2b) check if the first part is in the enabled languages

3) if - is not present lowercase it all
  3a) check if the whole text is in the enabled languages
```

NSkelsey:
The state machine for the regex that matches valid looking lang strings:

screen shot 2016-04-28 at 8 21 13 pm

Source: `/^([a-z]{2})(_[A-Z]{2})?$/`

NSkelsey:
### Still TODO

- Combine the 30 or so angular-i18n js files into the build.
- Resolve the bug mentioned in 7bd1a7a
- Discuss node request as an application wide promise.

Issue #1660, opened 2016-05-05
Users can bypass forced password change


fpietrosanti:
This ticket is to analyze a possible bug that has been reported by Faffa, related to the fact that a Recipient that should change a password (1st login), is able to bypass the password and goes straight the tip list.

It's a low priority bug, but to be consider as with client-side it become mandatory

evilaliv3:
yep this is something existing since ever and that i'm viewing to fix with @NSkelsey.

a user can currently navigate the pages without any check on that and he will get re-proposed the password change as he re-log in next times.

evilaliv3:
This will becoming more relevant with the end2end password key creation so that i will take care now of implementing it while completing it and eventually anticipating it to the master branch.

PR #1662, opened 2016-05-06
Fix issue #1660

evilaliv3:
Implemented but i'm not completely satisfied.

basically there is now a variable that if set in the rootscope allow to set the forced locations (i.e.: $rootScope.forcedLocation = '/antani') but we will have than some code here and there were to reset that variable (i.e.: delete $rootScope.forcedLocation).

@NSkelsey could you please see if you have some ideas to do it better and if not anyhow merge this?

### DIFF
--- a/client/app/js/controllers/main.js
+++ b/client/app/js/controllers/main.js
@@ -409,13 +409,6 @@ GLClient.controller('MainCtrl', ['$q', '$scope', '$rootScope', '$http', '$route'
 
     //////////////////////////////////////////////////////////////////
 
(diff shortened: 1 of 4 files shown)


NSkelsey:
@evilaliv3 looking at the source code I am – like you dissatisfied – with the implemented solution.

Using `$rootscope` here to communicate a variable related only to the authentication process is no good.

But! A solution has come to me. I will add a commit on this branch.

evilaliv3:
:+1: :+1:  waiting for that :)

NSkelsey:
Merged.

Issue #1663, opened 2016-05-07
Support configuration for deployment with reverse proxy on sub-path that does not handle URL rewriting


fpietrosanti:
This ticket is to support configuration for deployment with reverse proxy on sub-path that does not handle URL rewriting.

While working on testing deployment of OpenWhistleblowing behind Cisco ACE 4710 cluster handling HTTPS termination of existing anticorruption authority website, with reverse-proxy configured on the path www.website.it/submission/, the URL forwarding happens to be symmetric (/submission/ is forwarded to GlobaLeaks as /submission/), this trigger a 404 not found.

This ticket is to implement some kind of workaround that enable the configuration of a "PREFIX" that, if configured, enable the platform to also answer HTTP requests on /PREFIX/handler (or suggest another way to do it).

evilaliv3:
Documented: https://github.com/globaleaks/GlobaLeaks/wiki/Advanced-configuration

PR #1669, opened 2016-05-11
Implement feature #1663

evilaliv3:
This pull request implemente the requirements of #1663;

Being an advanced feaure it's configuration is currently allowed only inside the /etc/default/globaleaks.

Eventually at the next database change we could evaluate to add a configuration variable to allow this configuration through the admin interface.


evilaliv3:
@NSkelsey could you please take care of reviewing this?

PR #1708, opened 2016-06-26
Feature/api errors notification

evilaliv3:
@NSkelsey could you please review this?


NSkelsey:
After some discussion it seems that only watching `405` instead of also including static files misses with `404` codes is the best tradeoff for simplicity and utility.

:tada: #Ok2Merge 🎉

Issue #1731, opened 2016-07-19
The CSS customization should apply to public pages


NSkelsey:
An adopter needs to remove CSS that they have applied to every page in their interface. It needs to go on the public pages and that's it.

In order to handle this in a sane fashion and to not break the customizations of many other sites I propose the following:

Append CSS classes to the root tag:
```<div id="BodyDefault" data-ng-if="!Utils.isHomepage() || !node.custom_homepage">``` 

This way we can also solve issues like #1203 . . . and others

NSkelsey:
This needs to be documented in the [Customization Guide](https://github.com/globaleaks/GlobaLeaks/wiki/Basic-customization-guide) for this to close.

NSkelsey:
Documented.

Issue #1725, opened 2016-07-16
Urls including lang parameter should be retested and proper unit test implemented.


evilaliv3:
While working on the BustiaEtica project it has been noted that urls including ```lang``` parameter (e.g.  ```url?lang=ca```) appear sometimes to not working properly.

This ticket is to keep track of the retesting of the functionality of the feature.

As part of the ticket it would be important to write an end2end test for protecting against possible regressions in the feature.

evilaliv3:
A proper retesting should include:

- navigating urls on at least two languages; one of the two can be the default language.
- on this ourls spwcificlly the following checks should be done:
  - software texts should be properly internationalized
  - user defined texts should be properly internationalized

evilaliv3:
fix and regressions tests have been implenented and are now integrated in the devel branch.

the fix will be part of the upcoming release.

PR #1738, opened 2016-07-24
Proposed/v2.62.9

NSkelsey:
In order to get the needed features out for Bustia Ètica, I have cherry-picked the relevant commits from `devel` that address issues:

- #1725 
- #1731 

This is a quick release that will sit in front of v2.63.0. It gives us more time to look at issues that require a migration separately.


evilaliv3:
merged manually.

Issue #1676, opened 2016-05-24
Mark experimental admin features as experimental


NSkelsey:
It is important for the admin to see what features are considered experimental and what are for sure stable.

There are multiple ways to describe it in the UI, and the UX needs to be adjusted accordingly.

cc/ @fpietrosanti

NSkelsey:
After a brief [search](https://github.com/globaleaks/GlobaLeaks/search?l=html&q=enable_experimental_features&type=Code&utf8=%E2%9C%93) here are the experimental features (all of them are enabled client side:

- Tip Scoring
- Conditional submission steps
- The custodian account creation

fpietrosanti:
If we can have a boolean for each of them, to be activated, along with a description of what the feature is, it would be a nice catch. That way the admin will have all those feature disabled at upgrade, unless explicitly enabled?

evilaliv3:
i'm ok for having booleans assigned to features.
i would say that everything could become subject to a protection to a boolean or configuration but i would say that in order to easy the develpment we really would require this that is my preferite ticket since ever: https://github.com/globaleaks/GlobaLeaks/issues/999

by the end of the summer i would like to see in place:
- https://github.com/globaleaks/GlobaLeaks/issues/999
- the json refactor to specific l10n tables discussed with @NSkelsey 

this are the two refactors i aim that would make more simple the develpment and would solve the growing performances issues.

vodkina:
Could you please add estimate to this issue? Thanks

evilaliv3:
@NSkelsey as discussed a week ago, the single boolean with the experimental features from my point of view is becoming tricky for the user because when the user have to use simple one (e.g. the custodian or the conditianal questions) the user starts to unlock all as a whole.

instead we would have to simply have a panel for "Optional Platform Features" where the user with precise checkboxes will be able to select specifically a dropdown.

Issue #999, opened 2014-10-27
It should be possible to add node variables to the glbackend without modifying the db format


evilaliv3:
Currently everytime we need to add a config variable to the node table the node table must be edited.

This ticket is to keep track of all the changes needed in order to develop a different database structure that would permit to add new variables without the need to change the db format.

This is needed in order to speed up the development and reduce the error of db migrations.

NSkelsey:
The preference would be to implement this feature in a similar way to Firefox handles their config.

http://kb.mozillazine.org/About:config

evilaliv3:
@NSkelsey with those two commits the PoC of the db changes i've in mind is in place.

it implements what we discussed that is:
- a config table
- a localization table for the variables of the config model that will require the internationalization.

in addition to our discussion i've added a default attribute where to archiving/updating the default value that is from my point of view important in order when performing a migration through the various versions.

NSkelsey:
The above commit 5ca13eb is a refactor of non l10n system config. The goal is to integrate the l10n config first in #1739, then complete these changes, dropping the node and template tables completely.

vodkina:
Is it possibile to estimate this issue? Thanks

evilaliv3:
Implemented by @NSkelsey  and integrated by me in devel branch.

PR #1749, opened 2016-08-16
Refactor system configuration

NSkelsey:
These changes address issue #999 and provide a solution for #1676

Notably the `Node` and `Notification` tables are no more. 

The migrations are in place and from this point onward it is a bug if they do not function.

@evilaliv3 I consider these changes ready to go. Obviously the timing on when to introduce them is a different decision.


backend/globaleaks/models/__init__.py
```diff
@@ -104,16 +102,17 @@ def update(self, values=None):
...
-                if previous and isinstance(previous, dict):
-                    previous.update(value)
-                    setattr(self, k, previous)
-                else:
-                    setattr(self, k, value)
+                    if previous and isinstance(previous, dict):
+                        previous.update(value)
+                        setattr(self, k, previous)
+                    else:
+                        setattr(self, k, value)
```
NSkelsey: This needs to go.

vodkina:
Could you please add estimate to this issue? Thanks

backend/globaleaks/models/__init__.py
NSkelsey: removed in 79e3a5c

backend/globaleaks/db/sqlite.sql
```diff
@@ -68,9 +87,12 @@ CREATE TABLE context (
     enable_two_way_comments INTEGER NOT NULL,
     enable_two_way_messages INTEGER NOT NULL,
     enable_attachments INTEGER NOT NULL,
-    status_page_message BLOB NOT NULL,
     presentation_order INTEGER,
     show_receivers_in_alphabetical_order INTEGER NOT NULL,
+    name TEXT NOT NULL,
+    description TEXT NOT NULL,
+    recipients_clarification TEXT NOT NULL,
+    status_page_message TEXT NOT NULL,
```
NSkelsey: Bug, bug, bug

backend/globaleaks/db/sqlite.sql
```diff
@@ -10,8 +32,8 @@ CREATE TABLE user (
     deletable INTEGER NOT NULL,
     role TEXT NOT NULL CHECK (role IN ('admin', 'receiver', 'custodian')),
     state TEXT NOT NULL CHECK (state IN ('disabled', 'enabled')),
+    description TEXT NOT NULL,
```
NSkelsey: bug

backend/globaleaks/db/sqlite.sql
```diff
@@ -531,3 +416,4 @@ CREATE INDEX field__template_id_index ON field(template_id);
 CREATE INDEX step__questionnaire_id_index ON step(questionnaire_id);
 CREATE INDEX context_questionnaire_id_index ON context(questionnaire_id);
 CREATE INDEX fieldanswer__internaltip_id_index ON fieldanswer(internaltip_id);
+CREATE INDEX config_group_index ON config(var_group);
```
NSkelsey: Need to add another index for config_l10n

backend/globaleaks/handlers/admin/user.py
```diff
@@ -87,8 +87,6 @@ def create_receiver_user(store, request, language):
 
 
 def db_create_user(store, request, language):
-    fill_localized_keys(request, models.User.localized_keys, language)
-
```
evilaliv3: @NSkelsey: i think this has been removed for error, do you confirm?

backend/globaleaks/handlers/public.py
```diff
@@ -360,37 +327,35 @@ def _get_public_resources(store, language):
 
 
 class AhmiaDescriptionHandler(BaseHandler):
+    def _empt_gen(self):
```
evilaliv3: @NSkelsey when not needed let's use explicit identifier or common abbreviations.

e.g. empt in place of empty is a not needed optimization (save only one char) that simply make the code less naturally readable.

backend/globaleaks/handlers/public.py
```diff
@@ -360,37 +327,35 @@ def _get_public_resources(store, language):
...
-                                         serialize_node, self.request.language)
-
-        if node_info['ahmia']:
+        if GLSettings.memory_copy.ahmia:
             ret = yield GLApiCache.get('ahmia', self.request.language,
                                        serialize_ahmia, self.request.language)
 
             self.write(ret)
         else:  # in case of disabled option we return 404
+            yield self._empt_gen() # TODO remove inlineCallbacks from outer get.
```
evilaliv3: @NSkelsey: this can be simply replaced with "yield None" or even "yield" :)

Magic!

evilaliv3:
@NSkelsey: i applied really minor changes but i really enjoyed the rewrites you applied and the code quality :)

great pull request, well done!

PR #1824, opened 2016-12-05
Refactor build and init to exclude pyc files from packages and the file system

NSkelsey:
The newer versions of lintian consider shipping pyc files a [security issue](https://lintian.debian.org/tags/package-installs-python-bytecode.html). Currently globaleaks ships with all of its pyc files inside.

```bash
> lintian globaleaks_2.65.1_all.deb
E: globaleaks: package-installs-python-bytecode usr/lib/python2.7/dist-packages/globaleaks/__init__.pyc
E: globaleaks: package-installs-python-bytecode ... use --no-tag-display-limit to see all (or pipe to a file/program)
```

Using `PYTHONDONTWRITEBYTECODE` when app-armor is initialized preserves the read only rule for the python module directories. 

The negative is that this is a performance hit when using app armor, but how large of a performance hit is unclear. My intuition here is that the cost comes in more memory usage (to store the byte code) and at startup time (which is basically nothing).

### DIFF
--- a/backend/bin/globaleaks
+++ b/backend/bin/globaleaks
@@ -7,6 +7,9 @@ reload(sys)
 sys.setdefaultencoding('utf8')
 sys.excepthook = None
 
+# avoid to write bytecode files at runtime
+sys.dont_write_bytecode = True
+
 import os
 import random
 import re
--- a/debian/globaleaks.init
+++ b/debian/globaleaks.init
@@ -237,8 +237,6 @@ do_start()
 
     if [ "${APPARMOR_SANDBOXING}" -eq "1" ]; then
         APPARMOR_STARTED=0
-	# bytecode will not be written to .pyc files with app-armor enabled
-        export PYTHONDONTWRITEBYTECODE=1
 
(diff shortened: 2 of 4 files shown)


fpietrosanti:
Can we also document the security issue && good practice implemented somewhere in the security documentation?

evilaliv3:
Perfect analysis and patch Nick!

As small edit what we could do is to generate the conpiled files at
postinst time.
In fact apparmor apply only to the bin/globaleaks and not to all the rest.
This is also the recommended way by Debian guidelines: "Compiled Python
source files must not be included in the package. These files should be
removed from the package and created at package installation time in the
postinst."

As soon that i touch the laptop i will pass you the single line we used in
past to do that.

NSkelsey:
The BDFL confirmed in 2003 that there are no meaningful gains with pyc files. http://www.network-theory.co.uk/docs/pytut/CompiledPythonfiles.html.

For platform independence we can use: https://docs.python.org/2/library/sys.html#sys.dont_write_bytecode

Issue #1809, opened 2016-11-07
Officially support Ubuntu 16.04 LTS Xenial Xerus


NSkelsey:
This is a ticket to track requirements going into a PR for official support of 16.04.
The branch currently used is `testing/appliance-infra`

- [x] python package dependencies selected and frozen.

Plan to deploy using the 16.04 packages in Ubuntu's archives instead of using pip to handle the requirements at install time.

- [x] backend pip selection requirements relaxed
- [x] single command installation with `install-globaleaks.sh`
- [x] Review the debian package [upgrade guide](https://www.debian.org/doc/debian-policy/upgrading-checklist.html) and update accordingly
- [x] testing of installation on OS
- [x] debian package release and listed on deb.globaleaks.org

It is not clear exactly how to test 16.04 with current infrastructure. A packer build system could be set up in a private cloud to automate and test the build. Or more likely we will use local VMs until it becomes a blocking issue.

NSkelsey:
I believe with the non Xenial builds we may be violating a rule of the policy-checklist for upgrade to policy version `3.7.9`:

[For upgrade `3.7.9.0`](https://www.debian.org/doc/debian-policy/upgrading-checklist.html)
> Section 4.9:
> debian/rules: required targets must not attempt network access.

The older installation does not obey this rule since the package issues a pip request in the postinst step.

evilaliv3:
The packaging is accomplished, what is missing is final retesting and release that i will try to accomplish by the next week.

evilaliv3:
Support for Xenial is released and published on the repository.

Websites and wiki pages have been updated as well.

PR #1827, opened 2016-12-08
Add support for Xenial with install-globaleaks.sh

NSkelsey:
This PR tracks the changes to the build/packaging and related scripts to support Xenial 16.04.

There are some large changes in here:

0) The appliance changes here should produce a functional Xenial image using the instructions found in `appliance/Readme.md`

1) The Xenial package uses the dependencies in debian/controlX/control.xenial as the definition of its dependencies.

2) The requirements check for xenial requirements is performed with a `>=` instead of a hard equals at bootup. This allows the deps to shift underneath according to the updates of the system.

3) The apparmor profile was slightly expanded. `usr.bin.globaleaks` was asking to use `/proc/` with xenial so I added the read permission.

Some notes:

The entire process from the build to installation has been tested in local vms for both Xenial and Trusty, they both work start to finish. 

If the behavior of the trusty build changes due to changes introduced here it is a bug.

Related to #1809


debian/controlX/control.xenial
```diff
@@ -20,22 +20,21 @@ Depends:
...
+ , python-cryptography (>=1.2.3)
+ , python-cyclone (>=1.1)
+ , python-gnupg (>=0.3.8)
+ , python-openssl (>=0.15.1)
+ , python-scrypt (>=0.6.1)
+ , python-storm (>=0.19)
+ , python-transaction (>=1.4.3)
+ , python-twisted-core (>=16.0.0-1)
+ , python-txsocksx (>=1.15.0.2-1)
+ , ${python:Depends}
```
NSkelsey: @evilaliv3 `{python:Depends}` has snuck in again. The deb package as is installs without a hitch though.

NSkelsey:
@evilaliv3 this just needs a relabel to be integrated into devel

NSkelsey:
Will be merging this on 17/12

evilaliv3:
okay!

Issue #1831, opened 2016-12-15
Send all unhandled stack exception by default ALSO to website node administrator email


fpietrosanti:
This ticket is to send all unhandled stack exception by default ALSO to website node administrator email (if configured).

Actually we are receiving all unhandled stack exception but the website owners doesn't know if something is not working.

If the admin will receive the very same unhandled stack exception that we receive, will be stimoulated to upgrade the software to the latest version and/or to open up to us a support request.

NSkelsey:
This is a great idea. The stack exception mails should be configurabile as
a list of addresses where the admins email is added at startup.

We could add the admins mail in the startup wizard by adding the addresses
provided to the list.

Tangentially related to #1830
Il 15 dic 2016 10:20 AM, "Fabio (naif) Pietrosanti" <
notifications@github.com> ha scritto:

NSkelsey:
I started the implementation of this feature @evilaliv3. Two decisions were made that could go differently:

1. Only one mail will be sent for each exception to all addresses in the list.
2. There will only be one key that will be used to encrypt the exception notifications as they go out.

The goal here is to keep the changes small and keep the exception logic straight forward. Along with the UI!


<img width="515" alt="schermata 2016-12-23 alle 6 20 05 pm" src="https://cloud.githubusercontent.com/assets/1095794/21463806/c919398c-c93c-11e6-9026-69b53caf5692.png">

evilaliv3:
What about a boolean configuration in the admin users to mark if they will
have to receive the exceptions?

This way we can have a global setting with an enabler and a key + each
receiver with an enabler and it's key.

I would not go for a separated list/model just for this. What do you think?

evilaliv3:
Actually i like a lot that you this time proposed a small fix, but was not my intent to guide to you this approach aalways when designing new features, but just for bugfixing :)

What i dislike instead of this minimal fix is the fact that the PGP key will be the same for everyone.
Probably what i would do it to have:
1) The notification to us using a POST HTTPS request to a server of us (subject to an enabler that the admin could be eventually disable in the same way like one do for anonymous statistics in Chrome or other softwares)
2) Notification to the admin that will use naturally their key.

2) will be anyhow be necessary for implementing Capabilities for Admin/Custodians

@fpietrosanti ?

fpietrosanti:
This ticket is about implementing a 60 minutes fix, that's sending exception email by default also to configured admin email. 
[...]

PR #1851, opened 2017-01-03
Feature/1831

NSkelsey:
There are a few things left to do, but they could get done in devel:

- [x] ~~fix up email_addr_list_v or trust the request validator with the normal email regex.~~
- [x] upload send_exception.sh
- [x] ~~add a check for 2 email addresses in admin > notification > tab4 in e2e tests~~


evilaliv3:
@NSkelsey: as discussed please follow the directions for the implementation of this were written on https://github.com/globaleaks/GlobaLeaks/issues/1831

as suggested by @fpietrosanti the fix should be a 1 hour fix involving a simple for loop in sending the email that without any configuration should send the exception email to all the administrators.

this way still without any additional configuration the email notification will be encrypted with the specific admin configuration without requiring any additional implementation.

specifically all the code should be simply a for loop inside https://github.com/globaleaks/GlobaLeaks/blob/master/backend/globaleaks/utils/mailutils.py#L230 in function send_exception_email

backend/globaleaks/handlers/wizard.py
```diff
@@ -71,6 +71,10 @@ def wizard(store, request, language):
 
         db_create_admin_user(store, admin_dict, language)
 
+        # Add the new admin's email to list of recipients of errors
+        notif = NotificationFactory(store)
+        notif.get_val('exception_email_address_list').append(admin_dict['mail_address'])
```
evilaliv3: the reason why this would not work are:
- the admin email can change during time, so that this setting will work at initialization and will require manual management.
- the administrators can be more than
- existing nodes (that are the one from which we have more interest in notifying them the issues) will have to manually configure it.

NSkelsey:
Admin mails are now used and held in memory via GLSettings.memorycopy.notif

Issue #1858, opened 2017-01-07
Adjust GlobaLeaks Readme.md with new contact information


fpietrosanti:
This ticket is to adjust  GlobaLeaks Readme.md with new contact information https://github.com/globaleaks/GlobaLeaks#community-and-support following implementation of #1856

vodkina:
https://github.com/globaleaks/GlobaLeaks/pull/1887

PR #1887, opened 2017-01-27
Added new contact information

vodkina:
as per issue https://github.com/globaleaks/GlobaLeaks/issues/1858


NSkelsey:
Thanks for the pull request @vodkina!

We usually stage all of our changes in devel before pushing them into master so that the flow is one way.

I will add some edits here and merge the PR into devel.

PR #1896, opened 2017-02-08
Remove bower reducing client package manager to one: npm

evilaliv3:
@NSkelsey: here we go!


NSkelsey:
Ho paura ma andiamo avanti.

I would like to see the build work before we push this into devel!

evilaliv3:
time to merge!

PR #1935, opened 2017-03-27
Fix Markdown tables in README

EdOverflow:


evilaliv3:
Hey @EdOverflow!

Thanks for the contribution.

this fix is already present in devel since few days :) but thank you anyway

PR #1971, opened 2017-06-01
Cyclone replacement with minimal REST API interface

evilaliv3:
This pull request propose a cyclone library replacement with a minimalistic implementation of a REST API interface.

This is requirement for:
 - moving code to be more python3 compatibile
 - removing a library that is not anymore actively maintained
 - improving code safety adding some enforced constraints on the handler definition in order to enforce security of the API.

The implementation is a mixture of ideas inspired from python-klein that is currently not an option as it is not officially debian packaged.


NSkelsey:
@evilaliv3 please change the base of this branch to `devel`

PR #1990, opened 2017-06-30
Adopt torproject.org official repository for setup of Tor

evilaliv3:
@NSkelsey  could you please review this?

Code is fully based on past code: https://github.com/globaleaks/GlobaLeaks/commit/f549f55669769ece97a20fdab63e02b3391c3f8d
https://guides.github.com/features/mastering-markdown/

\cc @fpietrosanti


NSkelsey:
The motivation for this change is two fold. The first is that the latest versions of tor support the ADD_ONION tor control port commands and the second is that newer versions resolve AppArmor issues easier.

As verified Xenial supports v2.7.6 as of today which supports ADD_ONION.

```
  o Major features (controller):
    - Add the ADD_ONION and DEL_ONION commands that allow the creation
      and management of hidden services via the controller. Closes
      ticket 6411.
```
[source](https://raw.githubusercontent.com/torproject/tor/master/ChangeLog)

Second, according to trac.torproject.org ticket [#15588](https://trac.torproject.org/projects/tor/ticket/15588) handling of AppArmor specific configuration has been pushed down to distro maintainers as a packaging task because the core tor project team does not have the expertise.

NSkelsey:
Reopened for discussion because `v2.7.7` and onward contain security fixes that `v2.7.6` in Ubuntu universe does not support.

evilaliv3:
I propose to adopt >=0.2.9.11 for the following reasons:
- the first valid is 0.2.7.7 that include important security fixes.
- ubuntu xenial includes only 0.2.7.6

This will force us to use always the Tor repository (that could make sense for security reasons) but as @NSkelsey rightly points will limit to have a debian package included in official repositories for a while.
An option is to have just >=0.2.7.7 hoping that soon will be included in xenial (but anyhow we do not have our package included in xenail and we obviouslty wont and this is why i remain for the 0.2.9 that is the LTS).
In addition 0.2.7 will be officially deprecated by Tor on August 2017 and this give strenght to the point of moving to at least to 0.2.9.X

\cc @fpietrosanti

NSkelsey:
RE security fixes for v2.7.7: It is reasonable to expect that the tor package in the universe will be updated eventually or that one of use can become involved in ensuring it stays up to date.

RE using the official Tor PPA, requires existing nodes edit `/etc/apt/sources.list` or with these changes the `apt upgrade` will break for globaleaks because the packages  > v2.7.6 are not available in the standard repos.

evilaliv3:
I see the points and are legit.

but i do not see alternatives;
it is not acceptable in my poiint to keep a package that is declared by Tor to be not maintained starting from August 2017 and that is told by Tor to contains relevant security vulnerabilities with suggestion to move away from it.

NSkelsey:
Currently in Debian 9 stretch the [current Tor package](https://packages.debian.org/stretch/i386/tor) is `v2.9.11`. I believe we can issue a [SyncRequest](https://wiki.ubuntu.com/SyncRequestProcess) to the Ubuntu MOTU to push the Tor package forward to v2.7.7 or v2.9.x because upstream the package is updated.

PR #2059, opened 2017-08-22
Spelling: Internet

comradekingu:
_Public_ Internet is a bit redundant, but in the interest of making it really clear what is going on, so be it.


NSkelsey:
You sir, are very very correct about face palming over the phrase public Internet.

Although, I would rather go with "publicly reachable over the infotainment portion of the DARPAnet"

PR #2060, opened 2017-08-23
Spelling: No they, Via

comradekingu:


evilaliv3:
Thanks @comradekingu 

going to merge the patch but keeping "via Tor" lowercase for a while to not loose all the translations.

PR #2081, opened 2017-09-12
Feature/1727

NSkelsey:
These changes add full support for 1727. 

The end2end tests to test this particular functionality are missing, but would add complexity to the e2e tests for the admin-questionnaire setup and the submission testing.

Further the implementation could be refined to increase its performance when using many field-option triggers. However, that should be punted as the performance hit seems minimal.


evilaliv3:
Merged;

i've removed the need for checking for the key of the fields answers based on a regexp; that was not compatible with fields like 'whistleblower_identity' or other default fields that we will implement in the futur.

Let's see for the moment if we can avoid to have such a regexp check.

PR #2086, opened 2017-09-19
Feature/2019

NSkelsey:
This commit while missing the changes to the https_files_cfg.html string changes, is ready to go pending a review.

There are around 5 new strings added in this branch with nothing large on that side.

The end2end test for the wizard interface has also been updated, so this should pass travis testing, if only superficially because the embedded https ui remains untested.

### DIFF
--- a/client/tests/end2end/test-admin-perform-wizard.js
+++ b/client/tests/end2end/test-admin-perform-wizard.js
@@ -1,4 +1,4 @@
-describe('globaLeaks setup wizard', function() {
+fdescribe('globaLeaks setup wizard', function() {
   it('should allow the user to setup the wizard', function() {
     browser.get('/#/wizard');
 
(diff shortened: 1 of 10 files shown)


evilaliv3:
Picking up the review and integration in devel, with the goal of accomplishing this by the weekend.

client/tests/end2end/test-admin-perform-wizard.js
```diff
@@ -1,4 +1,4 @@
-describe('globaLeaks setup wizard', function() {
+fdescribe('globaLeaks setup wizard', function() {
```
evilaliv3: @NSkelsey: this would have probably to be reverted before integration?

evilaliv3:
Closing as outdated; Code will be eventualy integrated manually while reworking on this topic.

PR #2163, opened 2018-02-11
bug/652

vondrakk:
Added logic to provide user visual feedback (field error) when no checkbox is marked in a multiselect with required option set. Note that the submission step failed correctly without the change but the field was not highlighted as having an error state.


evilaliv3:
Thank you @vondrakk and sorry for delay in the feedback.

I think you patch is correct and i will soon retest it for integration.

PR #2224, opened 2018-03-19
First pass at implementing signup alert emails (fixes 2196)

NCommander:
Signed-off-by: Michael Casadevall <mcasadevall@ubuntu.com>


evilaliv3:
Great start @NCommander ! thank you!

evilaliv3:
@NCommander

I left you minor suggestions and some answer to your question on the ticket.

I've still not tried it but it look all right to me! well done!

NCommander:
Fixed the typo in tab2.

There are still some issues with this PR so I don't recommend merging it yet; you commented before I could update the bug but I think I need to pipe down the L10n factory + userdata to the templating layer as I ran into a case where we need to dynamically load a translation into an email when using use_data.

evilaliv3:
Integrated with very minor changes.

Thanks @NCommander !

PR #2239, opened 2018-03-29
Add translatable string "Surname"

### LINKED ISSUE #2238: Field "Surname" not translatable
https://try.globaleaks.org/#/signup

surname

comradekingu:
Fixes #2238


evilaliv3:
Thank you @comradekingu.

Actually the placeholder is already under internationalization but the string is not on transifex.

going to issue a patch slightly different.

PR #2280, opened 2018-05-12
Corrects str/bytes mismatch that prevents Py3 file uploads from working

NCommander:


evilaliv3:
well done @NCommander !

PR #2301, opened 2018-05-30
Questionnaire Duplication

NCommander:
Implements the ability to duplicate questionnaire's in the UI.


evilaliv3:
Thanks @NCommander, the patch looks good!

Could you please apply the following two changes:
- field triggers serialization is broken and requires ID fixing.
- the clone+insert_name+save would benefit to be implemented as a modal; as a start we we may reuse code of Tip deletion that implement similar logics)

NCommander:
I reworked this to fixed trigger fields and it appears to work correctly. I'm not entirely sure all the options are handling propery as I can't find the definitive list of how those dicts work but it should be OK now.

evilaliv3:
Thanks @NCommander, the corrections are in a good shape.

I just fixed your algorithm to support recursion as a field may contain fields recursively.

Issue #2317, opened 2018-06-13
GlobaLeaks should gracefully handle ACME server failures when no JSON response occurs


fpietrosanti:
Platform:  
Host:  
Version: 3.1.8

ValueError Inappropriate argument value (of correct type).

Traceback (most recent call last):

  File "/usr/lib/python2.7/dist-packages/globaleaks/jobs/base.py", line 53, in run
    yield self.operation()

  File "/usr/lib/python2.7/dist-packages/twisted/internet/defer.py", line 1126, in _inlineCallbacks
    result = result.throwExceptionIntoGenerator(g)

  File "/usr/lib/python2.7/dist-packages/twisted/python/failure.py", line 389, in throwExceptionIntoGenerator
    return g.throw(self.type, self.value, self.tb)

  File "/usr/lib/python2.7/dist-packages/globaleaks/jobs/certificate_check.py", line 75, in operation
    yield self.check_tenants_for_cert_expiration()

  File "/usr/lib/python2.7/dist-packages/twisted/python/threadpool.py", line 246, in inContext
    result = inContext.theWork()

  File "/usr/lib/python2.7/dist-packages/twisted/python/threadpool.py", line 262, in <lambda>
    inContext.theWork = lambda: context.call(ctx, func, *args, **kw)

  File "/usr/lib/python2.7/dist-packages/twisted/python/context.py", line 118, in callWithContext
    return self.currentContext().callWithContext(ctx, func, *args, **kw)

  File "/usr/lib/python2.7/dist-packages/twisted/python/context.py", line 81, in callWithContext
    return func(*args,**kw)

  File "/usr/lib/python2.7/dist-packages/globaleaks/orm.py", line 108, in _wrap
    result = function(self.instance, session, *args, **kwargs)

  File "/usr/lib/python2.7/dist-packages/globaleaks/jobs/certificate_check.py", line 43, in check_tenants_for_cert_expiration
    self.cert_expiration_checks(session, tenant[0])

  File "/usr/lib/python2.7/dist-packages/globaleaks/jobs/certificate_check.py", line 57, in cert_expiration_checks
    db_acme_cert_issuance(session, tid)

  File "/usr/lib/python2.7/dist-packages/globaleaks/handlers/admin/https.py", line 518, in db_acme_cert_issuance
    Settings.acme_directory_url)

  File "/usr/lib/python2.7/dist-packages/globaleaks/utils/letsencrypt.py", line 50, in run_acme_reg_to_finish
[...]

PR #2318, opened 2018-06-14
Change acme checkout behavior

NCommander:
Researched and looked deeply at  #2317.

Fundamental issue that causes the bug is that we re-raise on ACME failure (though we probably shouldn't be trying to catch Exception). I think the correct course here is to simply keep trying to renew the certficate from 30->15 days, but I want to check with the LE guys to make sure we don't accidently run afoul the rate limits or similar. Specifically, we should howler if we get validation failures, otherwise, just keep retrying. I'll push a better PR on that as soon as I figure out the specifics.


NCommander:
@evilaliv3 this is ready for merge after review

evilaliv3:
Thank you @NCommander!

I'm picking up this review where i'm going to apply the following changes:
- simplify texts for the user email that are a bit too verbose with no particular added value for the user that would anyhow not have any possibility to fix it manually; In short the user should just know "Failure in renewal of the certificate"
- split the patch in multiple commits dividing at least code, from src texts, from data update
- Integration will be postponed after testing; Specifically for a feature like these that is deeply used we could not ship it without proper testing.

PR #2326, opened 2018-06-18
Fix form field input strings. Closes #2325

### LINKED ISSUE #2325: "This field is too short" is shown before entering the text while it's not shown that's mandatory
**Current behavior**
When a text field have a minimum length, the error message "This field is too short" is shown before entering any text, while the fact that the field is mandatory is not shown.

Reported by a customer on:
screen shot 2018-06-18 at 14 04 28

**Expected behavior**
While the field is empty and it's market as mandatory, it must display that the field is mandatory (information before entering any input), while it must display "the field is too short" after writing input on it.

NCommander:
Fixes #2325


evilaliv3:
Works great @NCommander !

I just made a change to make the error be read if validation of the full questionnaire is triggered.

Soon or later this part of the application would require major work in order to minimize the number of functions calls while the user type that causes some minor performance freeze on complex questionnaires.

Issue #2193, opened 2018-03-04
UX & Backend bug in Certificate Enrolment trigger: EnvironmentError: Response unexpected on verify_hostname


fpietrosanti:
Platform: 
Version: 3.0.7

EnvironmentError Base class for I/O related errors.

Traceback (most recent call last):

 File “/usr/lib/python2.7/dist-packages/twisted/internet/defer.py”, line 1128, in _inlineCallbacks
   result = g.send(result)

 File “/usr/lib/python2.7/dist-packages/globaleaks/handlers/admin/config.py”, line 78, in verify_hostname
   raise e

EnvironmentError: Response unexpected

fpietrosanti:
This bug happen because:
- Backend does not verify that hostname is null
- Client does allow to write an hostname that's not saved, but user click as action "generate" the certificate. 

The fix should be on the backend integrity check and on the client, skipping the need to save manually, if a new hostname has been typed in.

DavidLemayian:
I'm getting this error when trying to verify hostname. Blocking SSL certificate renewal. 🤔

evilaliv3:
@DavidLemayian: which version are you running?
We had an issue with Letsencrypt renewal that should be now solved starting from release 3.10.6

p.s. the update will solve the renewal issue but probably wont solve the verify hostname defect that is caused by a different issue still to be identified.

DavidLemayian:
Makes sense. Upgrade solved this. Thanks @evilaliv3 .

evilaliv3:
Closing as fixed.

PR #2331, opened 2018-06-19
Forcibly save the hostname when generating an SSL certificate

NCommander:
Signed-off-by: Michael Casadevall <michael@casadevall.pro>


NCommander:
Resolves #2193

PR #2353, opened 2018-07-06
UserTenant endpoint/UI

NCommander:


evilaliv3:
@NCommander: i've reviewed the pull request, good work!

Few notes on the correction applied:
- made serialize_tenant_association to just serialize an UserTenant row with no joins with Tenant table; this is for having simple API and will be the client to merge objects when necessary. this would remove many overhead and increase backend performances when the overhead on the client will be minimal.
- fixed a security issue on the UserTenant handler that was missing 'user_tenant_only' flag; this would have lead anybody to be able to associate a user to anybody else site.
- removed (at least for the moment) UserTenantCollection.get() that is never used.
- reordered functions definition in bottom-up approach from the inner functions to the macro components so to ease readability
- reordered imports and removed unused imports probably

Issue #2337, opened 2018-06-26
Make it possible to reset the onion service key via the web interface


evilaliv3:
Many projects when cloning an existing platform in order to develop an infrastructure with a pre-production and production environment require to have two nodes with different hidden services.

This is currently only possible via the following command line:```gl-admin setvar tor_onion_key ""``` that make it possible to reset the existing onion service key; a restart of the application is required to trigger the new generation.

This ticket is to keep track of the work for making it possible to reset the onion service key via the web interface.

The interface, near the hidden service name may show a refresh icon with the possibility of re-generating the hidden service key.

When pressed the button should just reset the existing tor_onion_key variable and trigger the regeneration already implemented inside the onion service job.

evilaliv3:
I've analyzed this feature for the necessity of any db changes and actually it would not require any change.

@NCommander: Implementation for this feature could be probably easily added via an operationHandler command reset_onion_key to be bound to a button near the onion service name?

PR #2356, opened 2018-07-11
Implement UI method to reset Tor Onion Service (Closes: #2337)

NCommander:
Signed-off-by: Michael Casadevall <mcasadevall@ubuntu.com>

### DIFF
--- a/backend/globaleaks/handlers/admin/operation.py
+++ b/backend/globaleaks/handlers/admin/operation.py
@@ -91,7 +91,7 @@ class AdminOperationHandler(OperationHandler):
     @inlineCallbacks
     def reset_onion_private_key(self, req_args, *args, **kargs):
         yield set_onion_service_info(self.request.tid, None, None)
-        yield State.onion_service_job.add_hidden_service(None, None, self.request.tid)
+        yield State.onion_service_job.add_hidden_service(self.request.tid, None, None)
         yield State.onion_service_job.remove_unwanted_hidden_services()
 
         # Return the new key
--- a/backend/globaleaks/handlers/admin/operation.py
+++ b/backend/globaleaks/handlers/admin/operation.py
@@ -9,7 +9,9 @@ from globaleaks.models import Config
 from globaleaks.models.config import ConfigFactory
 from globaleaks.orm import transact
 from globaleaks.rest import errors
+from globaleaks.state import State
 from globaleaks.utils.utility import is_common_net_error
+from globaleaks.jobs.onion_service import set_onion_service_info, get_onion_service_info
 
 from six import text_type, binary_type
 from six.moves.urllib.parse import urlparse, urlunsplit # pylint: disable=import-error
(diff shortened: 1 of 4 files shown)


backend/globaleaks/handlers/admin/operation.py
```diff
@@ -9,7 +9,9 @@
 from globaleaks.models.config import ConfigFactory
 from globaleaks.orm import transact
 from globaleaks.rest import errors
+from globaleaks.state import State
```
evilaliv3: @NCommander: state is directly accessible from handlers and jobs using self.state.

As well in case Settings would be required it would be accessible via self.state.settings

Going to apply myself a fix as part of the integration review

client/app/views/admin/network/tor.html
```diff
@@ -4,10 +4,17 @@
...
-                  <span ng-hide="admin.node.onionservice === ''"  class="label label-primary">{{admin.node.onionservice}}</span>
-                  <span ng-show="admin.node.onionservice === ''" class="label label-warning" data-translate>Not running</span>
-              </label>
+            <div ng-hide="admin.node.onionservice === ''">
+              <p class="input-group">
+                <span class="input-group-addon" data-translate>Tor Onion Service</span>
+                <input name="onionservice" class="form-control" type="text" data-ng-model="admin.node.onionservice" data-ng-disabled="true">
+                </input>
+                <span class="input-group-btn">
+                  <button class="btn btn-danger" data-ng-disabled="torOnionResetInProgress === true" data-ng-click="resetOnionPrivateKey()" data-translate>Regenerate</button>
```
evilaliv3: no need for === true on booleans

client/app/views/admin/network/tor.html
```diff
@@ -4,10 +4,17 @@
       <div class="row">
         <div class="col-md-12">
           <div class="form-group">
-              <label><span data-translate>Tor Onion Service</span>:
-                  <span ng-hide="admin.node.onionservice === ''"  class="label label-primary">{{admin.node.onionservice}}</span>
-                  <span ng-show="admin.node.onionservice === ''" class="label label-warning" data-translate>Not running</span>
-              </label>
+            <div ng-hide="admin.node.onionservice === ''">
```
evilaliv3: missing 'data-'

in addition we may use a simple data-ng-show="admin.node.onionservice"

no need for === ' '  on empty strings, we can use

client/app/views/admin/network/tor.html
```diff
@@ -4,10 +4,17 @@
...
+              <p class="input-group">
+                <span class="input-group-addon" data-translate>Tor Onion Service</span>
+                <input name="onionservice" class="form-control" type="text" data-ng-model="admin.node.onionservice" data-ng-disabled="true">
+                </input>
```
evilaliv3: input tag is self closing

PR #2382, opened 2018-07-30
Fix salt to be consistent on Py2/Py3.

NCommander:
Signed-off-by: Michael Casadevall <michael@casadevall.pro>

This likely needs a migration fix to go with it. As far as I can tell, we don't use salt anywhere ATM, so maybe we can blow it away and just regenerate?


evilaliv3:
@NCommander: migration should not be necessary and would break the receipts that are using it because if we change the salt the hash would not match.

I think we could just keeping it as is as it is only used for password hashing and a more bigger payload would not cause any issue hopefully.

PR #2405, opened 2018-09-12
Crypto

evilaliv3:
This pull request implement changes documented in https://docs.google.com/document/d/1Yn4OM5XO5G0PXDSlHYaEa5BsAucBmxhmBfRqhrAg_Jc/edit#heading=h.pfeyp5s869i5

Main commit of the pull request is: https://github.com/globaleaks/GlobaLeaks/commit/11235de70d435fbc22c1eaec9ebc64fa5dfb24ed


evilaliv3:
@NCommander: When you like can you review this pull request and in particular commit https://github.com/globaleaks/GlobaLeaks/commit/151eddb88ad7c393c4307a7d52667a40b044ce35?

All should be ready for integration (aside from bugfixing), I've to accomplish only the streaming decryption but the commit is on its way.

\cc @fpietrosanti

evilaliv3:
@NCommander we discussed it all offline but here the answers for other readers:

1. argon2 is our preference since we evaluated KDF but was not adopted because was not available. the choice for it is that it enables to decouple RAM requirements from time requirements.
While with scrypt with 128MB of ram we would have a small time requirement in terms of milliseconds, with argon2 we can set it to 1second per login. Will this defeat bruteforce even on receipts that are 10^16 because if you ask 1 second with no parallelization or with the parallelization possible requiring 128MB will be unfeasible to bruteforce that space.

2. aside from bugs the code is written to work with version of pynacl lower than 1.2 where simply encryption wont be available. the reason why in pip requirements we will leave 1.2 for the moment is that unit tests expect encryption to be available. (except in some tests were it is set to be disabled explicitly)

3.
- we use always asymmetric when encrypting tips, comments, files, messages, and the private key of the tip
- we use symmetric for encrypting the private key of the user

The reason why tips,comments,files,messages are encrypted using asymmetric is that it enables us to perform it in a job / async way because the public key is stored on the db and available to jobs. This will eventually make it possible to even encrypt old tips in a batch job as soon that keys has been generated.

PR #2406, opened 2018-09-13
Import export rebase

evilaliv3:


backend/globaleaks/utils/import_export.py
```diff
@@ -0,0 +1,419 @@
...
+                            if file_obj.status == u'reference':
+                                print("Reference file " + file_to_read + " not found, skipping.")
+
+            process_files_for_tarball(tenant_data['receiverfile'])
+            process_files_for_tarball(tenant_data['whistleblowerfile'])
+
+    finally:
+        shutil.rmtree(dirpath)
+
+    return output_file.getvalue()
```
evilaliv3: @NCommander:
This function is correct but the specific handling of this getvalue and of the Bytes buffer make it write it all in RAM.
Use a real file in place of a RAM buffer and return the file handle.

client/app/js/controllers/admin/tenant.js
```diff
@@ -12,12 +12,22 @@ angular.module('GLClient')
...
-  }
+  };
+
+  $scope.importTenant = function(file) {
+    $http({
+      method: 'POST',
+      url: 'admin/tenants/import',
+      data: file,
+    }).then(function() {
+       $route.reload();
```
evilaliv3: $route is not defined!

backend/globaleaks/handlers/admin/tenant.py
```diff
@@ -193,3 +201,18 @@ def delete(self, tenant_id):
...
+    root_tenant_only = True
+    invalidate_tenant_states = True
+
+    def get(self, tenant_id):
+        tenant_id = int(tenant_id)
+
+        return export_tenant(tenant_id)
+
+    def post(self):
+        import_tenant(self.request.content.read())
```
evilaliv3: @NCommander: this need to be handlerd  a file handler where the same way handlers/admin/file.py works.

Here you have a 1GB+ payload hitting you and you are reading it all in RAM.
With a file handler instead the file will arrive in streaming, and here will have just a file descriptor to pass directly to the import_tenant function;
Internally to the import_tenant function we may use the Tar functions to still read it in streaming writing unpacked data on the disk.

evilaliv3:
@NCommander:  when you will rework the streaming aspects and retest the patch, it would be really valuable if you could implement a basic unit test for the handler.

Already in unit tests you find some stub able to create multiple tenants (i think three are generated right now), we can generate them deterministically, export, reimport and check some values in different places just to have a basic verification on some variables here and there of diffeent type
with this approach, even if the check is not complete we could have basic CI testing useful to spot bugs as we will change things during time.

backend/globaleaks/utils/import_export.py
```diff
@@ -0,0 +1,419 @@
...
+    '''Collects the tenant data into dictionary form ready for serialization
+       into a new database instance detaching the information from SQLAlchemy'''
+
+    tenant_data = {}
+
+    # First recover the base tenant data
+    tenant = session.query(models.Tenant).filter_by(id=tid).first()
+
+    if tid == EXPORTED_TENANT_ID:
+        print("Importing Tenant: " + tenant.label)
```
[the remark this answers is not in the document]
NCommander: I vaguely remember there was a problem with using log when it's called from the context of gl-admin, but I'm not sure that's actually a problem. I'll fix this.

PR #2455, opened 2018-11-14
Fix #2454 

### LINKED ISSUE #2454: Username in the UserBox aren't the one of the logged-in user
**Current behavior**
The username show in the UserBox (in the right-high corner) doesn't update her value after log-in with different users.
The case that different users use the same computer and browser is a border line use case.

**Expected behavior**
The username showed in the UserBox are the one of the loggedin user.

**Steps to reproduce the problem or feature illustration**

1. Log-in with an user (p.e. admin).
2. See Admin in the UserBox
3. Log-out
4. Log-in with a different user (p.e. recipient).
5. See Admin in the UserBox

**What is the motivation or use case for changing the behavior?**
To show in the UserBox one name of the logged-in user.

**GlobaLeaks version:**
v3.4.1
devel

**Browser:**
Chrome 70
Firefox

**Server Operating System and Version (if applicable):**
Ubuntu(18.04)

**Client Operating System and Version (if applicable):**
Ubuntu(18.04)

davelip:
#2454 Username are not the one of logged-in user.

Removed one-time binding on the preferences.name variable

### DIFF
--- a/client/app/views/partials/user.html
+++ b/client/app/views/partials/user.html
@@ -7,7 +7,7 @@
   <div id="LoginStatusBox" class="row" data-ng-if="Utils.showUserStatusBox()">
     <div class="pull-right">
       <span id="UsernameBox" data-ng-if="session.role !== 'whistleblower'">
-        <label>{{::preferences.name}}</label>
+        <label>{{preferences.name}}</label>
         <span class="text-separator">|</span>
       </span>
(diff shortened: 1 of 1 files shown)


evilaliv3:
Thank you @davelip, the patch is correct and useful!

Going to merge it and will be released in the upcoming 3.5.0

Issue #2463, opened 2018-11-26
GL 3.5.5 the flag visibility_subject_to_authorization in whistleblower_identity field are not saved/retrieved correctly from the database


davelip:
**Current behavior**
In /#/admin/questionnaires the standard field whistleblower_identity (in IT "Vuoi fornire i tuoi dati identificativi?") has a field visibility_subject_to_authorization (in IT "Assoggetta la visibilità della risposta all'autorizzazione da parte del custode"), the choise done over that checkbox doesn't are replied on the database, the value in the database aren't showed on that checkbox. 
Same on prompt_confirmation_upon_negative_answer (in IT "Chiedi conferma in caso di risposta negativa").

**Expected behavior**
Changes on the interface are reflected on the database.

**Steps to reproduce the problem or feature illustration**
With an admin profile go to /#/admin/questionnaires add a standard whistleblower_identity field to a questionnaire, set to true visibility_subject_to_authorization and prompt_confirmation_upon_negative_answer.
Reload the questionnaire, your choise are lost.

**What is the motivation or use case for changing the behavior?**
It's a bug, in GL 3.4.1 this works correctly.

**GlobaLeaks version:**
GL 3.5.5

**Browser:**
Chrome 70

**Server Operating System and Version (if applicable):**
Ubuntu 18.04

**Client Operating System and Version (if applicable):**
Ubuntu 18.04

evilaliv3:
Thank you for reporting this @davelip.

We are investigating the issue.

evilaliv3:
Closing as fixed, will be released in 3.5.6!

evilaliv3:
Thank you @davelip !

If integrated your patch in place of mine to preserve the credits for your contribution: https://github.com/globaleaks/GlobaLeaks/pull/2464

PR #2464, opened 2018-11-27
After migration 45 fieldattr bool reads always to False

davelip:
Fix for #2463

backend/globaleaks/db/migrations/update_45/__init__.py
```
def migrate_FieldAttr
[...]
            if new_obj.type == 'bool':
                new_obj.value = new_obj.value == u'True'
[...]
```

With this bool value became really boolean

### DIFF
--- a/backend/globaleaks/handlers/public.py
+++ b/backend/globaleaks/handlers/public.py
@@ -242,9 +242,7 @@ def serialize_field_attr(attr, language):
         'value': attr.value
     }
 
-    if attr.type == 'bool':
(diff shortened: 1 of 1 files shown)


evilaliv3:
Thank you so much @davelip!

The correct patch is https://github.com/globaleaks/GlobaLeaks/commit/a6f29d994e0f64e61282cce7d8cf14363f51ac01 but your input was really helpful to identify the reason of the failure!

evilaliv3:
Thank you @davelip 

Actually this patch is not a fix but the removal of a piece of code really necessary.

Issue #2467, opened 2018-11-28
Hide the button "Request access to the whistleblower's identity" when the identity is undeclared


davelip:
**Current behavior**
In receiver's whistleblower identity view (in tip detail) the button "Request access to the whistleblower's identity" are always visible, also when that is undeclared.

**Expected behavior**
When WB identity are not declared receivers shouldn't view the button "Request access to the whistleblower's identity"

**Steps to reproduce the problem or feature illustration**
Open a tip in which the wb identity is undeclared, the button "Request access to the whistleblower's identity" is visible under the string "Undeclared"

**What is the motivation or use case for changing the behavior?**
It's inconsistent.

**GlobaLeaks version:**
3.5.5

**Browser:**
Chrome 70

**Server Operating System and Version (if applicable):**
Ubuntu 18.04

**Client Operating System and Version (if applicable):**
Ubuntu 18.04

evilaliv3:
Thank you @davelip!

The issue was legit and your patch correct.
Your code has been integrated in release 3.5.6

PR #2468, opened 2018-11-28
Hide Request WB identity button when its undeclared

davelip:
issue #2467


evilaliv3:
Thank you @davelip, the patch is correct and has been integrated in release 3.5.6 with very minor corrections!

PR #2480, opened 2018-12-31
Docker support

gionn:


evilaliv3:
Good morning @gionn and good year!

Thank you for your contribution! :)

I performed the review and here some considerations:
1. nightly: current packaging fetch the package from the globaleaks repository so that the container created always uses the same package (the latest on the repository); in case of a nightly we should probably use the the package built on travis.
2. release: the tag does generally happen before the publishing of the deb package, so that it is probably not possbile to publish the docker image on commit; i think the best would be to disable the push on travis and just perform it manully as part of the release procedure.
3. the current container seems to work well but it currently miss the startup of Tor; @ckeeney got in the same issue and solved it with a script (https://github.com/ckeeney/docker-globaleaks/blob/master/docker-entrypoint.sh) that causes docker to start correctly but has failures in shutdown.

@ckeeney @n-daniel do you have any advice?

gionn:
> 1. nightly: current packaging fetch the package from the globaleaks repository so that the  container created always uses the same package (the latest on the repository); in case of a nightly  we should probably use the the package built on travis.

Yeah, I can simply chain my docker scripts after the .deb creation job, and it should be good.

> 2. release: the tag does generally happen before the publishing of the deb package, so that it is probably not possbile to publish the docker image on commit; i think the best would be to disable the push on travis and just perform it manully as part of the release procedure.

we need to have a green build before pushing the new docker image so it should be done in the after_success step. Not sure if there is an easy way to grab the latest .deb built on the previous step ([travis provides artifact upload only on user-supplied S3](https://docs.travis-ci.com/user/uploading-artifacts/)), rebuilding again is ugly but fine if package generation doesn't depends on external factors.

> 3. the current container seems to work well but it currently miss the startup of Tor; @ckeeney got in the same issue and solved it with a script (https://github.com/ckeeney/docker-globaleaks/blob/master/docker-entrypoint.sh) that causes docker to start correctly but has failures in shutdown.

for many reasons it's better to use a separated image instead of doing multiple process management within the same container, I can provide a docker-compose file that is the common way to publish multiple containers for a single project. It's fine for globaleaks to have tor running on a separated node (reachable via 9050/tcp of course)?

evilaliv3:
Thank you for the prompt feedback @gionn!

1) good to go for such an edit; do you want to guide me through it?
the build and install job does currently generate the deb, create a a local repo and actually just perform apt-get install globaleaks (by using the regular install script)

2) release
for the moment i would suggest to skip releasing as the official packages are performed manually with no travis assistance; they in fact requires repository signing with offline keys.

3) actually not.
GlobaLeaks used tor as a network proxy but also for creating hidden services.
Currently for simplicity of the deployment it uses the default Tor Control Port as Unix Socket where GlobaLeaks access is granted by adding the globaleaks user to debian-tor group.

The alternative is to automate the creation of a shared secret and configure it in /run/tor/control.authcookie on the Tor VM and inside GlobaLeaks. This alternative requires to be able to make this configuration happen the first time the user boot the VM and obviously not at the time of the generation of the VM.

A quick and dirty trick would probably be to configure Tor for a passwordless control port (that is just a static config in /etc/torrc) port and make that port reachable only by the GlobaLeaks vm. As an As lternative 

@adrelanos: have you ever managed aspects like these in Whoonix?

fpietrosanti:
I feel that the self-contained design of GlobaLeaks, with the simplification of installation and maintenance, should also be followed for any further release system, in that case being Docker.

For that design goal, also as a matter of possible being able to spinoff GlobaLeaks instances at Digital Ocean and many other Docker-ready systems, the Tor application should be part of the single container, adjusting and/or modularising and/or adding new startup/shutdown scripts when running from within a Docker container (also considering that we are not using it, actually, so the more is integrated, the better will be maintained as it become used without code duplication).

evilaliv3:
Yep i agree actually.

@gionn what do you recommend?

gionn:
>        GlobaLeaks used tor as a network proxy but also for creating hidden services.
>        Currently for simplicity of the deployment it uses the default Tor Control Port as Unix Socket where GlobaLeaks access is granted by adding the globaleaks user to debian-tor group.

Control Port can be exposed as TCP, authentication is not a problem because the port will be available only from globaleaks container.

Howoever since it's not supported, I've pushed a Dockerfile with support for running tor alongside globaleaks via supervisord.

PR #2546, opened 2019-04-08
Fix YAML syntax in publiccode.yml

alranel:
I'm very sorry about the noise. My previous PR contained a YAML syntax error. This should be the last one.
Pros and cons of being early adopters :-)
Thank you for your patience.


evilaliv3:
Thank you @alranel !

alranel:
I'm afraid this PR didn't get through, as it does not show in the file history:
https://github.com/globaleaks/GlobaLeaks/commits/master/publiccode.yml

Maybe someone force-pushed and re-wrote the history of this repo?

PR #2560, opened 2019-04-23
Fix YAML syntax in publiccode.yml

alranel:
I'm afraid [my last pull request](https://github.com/globaleaks/GlobaLeaks/pull/2546) was overwritten by some force-push you did, because despite it was merged I can't see those fixes applied in master.
This pull request reapplies those lost fixes. Thank you for your patience!


evilaliv3:
Thank you @alranel, going to reapply it now!

Is there any update on the public visibility of the data?

PR #2590, opened 2019-06-18
fixed the build parameters

DhirajVarma:


evilaliv3:
Thank you @DhirajVarma;

Would you please clarify around your patch? There is no need for an argument in parameter ```n```

PR #2886, opened 2020-08-17
Use proper BCP 47 language tags in publiccode.yml.

bfabio:


evilaliv3:
Thank you @bfabio !

How did you discovered that we were failing in this? What could we keep as reference for this aspects?

bfabio:
Hello @evilaliv3, you're welcome!

I was investigating for the best way to fix this https://github.com/italia/developers-italia-backend/issues/162 and I wondered how [availableLanguages where defined in the standard](https://docs.italia.it/italia/developers-italia/publiccodeyml-en/en/master/schema.core.html#section-localisation).

In general, you could integrate the validator in your CI pipeline
```
docker run -i italia/publiccode-parser-go /dev/stdin < publiccode.yml
```
but it wouldn't have caught this one! (https://github.com/italia/publiccode-parser-go/issues/47)

PR #2961, opened 2021-01-13
Fix admin table cell widths

indigane:
Table cells in two admin views, Text customization and URL redirects, are broken due to bootstrap column classes on `th` heading cells.

These classes do not work in this context.

Instead, when using Chrome, the only effect that is applied is `width: 100%;` which causes the first cell to be as wide as possible, and the rest as small as possible, which along with the `* { word-break: break-word; }` global style makes the cells one character wide.

Firefox seems to handle 100% width on all cells differently, and instead distributes the space evenly.

### DIFF
--- a/client/app/views/admin/content/tab5.html
+++ b/client/app/views/admin/content/tab5.html
@@ -34,9 +34,9 @@
       <table class="table table-striped">
         <thead>
           <tr>
-            <th class="col-md-3" data-translate>Original text</th>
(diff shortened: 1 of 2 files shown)


evilaliv3:
Thank you for this correction @indigane 

Feel free to send all your suggestions.

I invite you to join us on slack at https://slack.globaleaks.org

PR #2962, opened 2021-01-13
Fix admin redirects table data

indigane:
This was changed from `path1` `path2` to `path2` `path3` in this commit https://github.com/globaleaks/GlobaLeaks/commit/10d4114b221b0ba518e185d717073ef5f53c8966#diff-0dc7bdfe6f435431b4fb5c703652258f46c11e8b11232850b3811ab96701890f

The effect is that "To" path ends up in "From" column in the table, and "To" column in the table is empty.

### DIFF
--- a/client/app/views/admin/network/url_redirects.html
+++ b/client/app/views/admin/network/url_redirects.html
@@ -35,9 +35,9 @@
         </tr>
       </thead>
       <tbody>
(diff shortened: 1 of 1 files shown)


evilaliv3:
Thank you for this correction @indigane

Feel free to send all your suggestions.

I invite you to join us on slack at https://slack.globaleaks.org

PR #2968, opened 2021-02-07
Friendly contact information

comradekingu:


evilaliv3:
@comradekingu: Would you please explain the reasons of your pull request?

comradekingu:
Certainly, https://github.com/fmeurou/geocurrency/pull/10#issuecomment-769710082

The usual nobody-asked-contributors doesn't help, it doesn't do anything well, if it supposedly does that could and is done anyway, and it has a lot of potential for harm.

Nowhere has it ever racked up anything but failures. It is bloat. Think of how much you can remember, and then tell me anyone can read that and hope to find people that don't read and are anti-social will somehow not be due to also disobeying a textfile. It is preposterous.

If you will, the Hosted Weblate community with almost 30k registered people did just fine without, and it was removed also without issue. Here it is from memory.

Welcome to the ___ community.
Accept and offer criticism constructively. Let anyone have their desired privacy.
Settle your differences between yourselves.
Finding yourself unable to do so, contact ___   title.

PR #3003, opened 2021-04-30
Docs: Glossary reworked

comradekingu:

- [ ] Your code should contain tests relevant for the problem you are solving
- [ ] A description of the problem you're trying to solve
- [ ] If the feature changes current behavior, reasons why your solution is better.

x


documentation/gettingstarted/Glossary.rst
```diff
@@ -1,34 +1,36 @@
...
+        An access token (effectively a password) generated by the *platform* and sent to the *whistleblower* to authenticate upon further visits to the *platform* after having completed a *submission*.
+With it, the *whistleblower* can modify and update their *submission* by uploading new files, adding comments to the *tip*, and read *recipients*'s comments about the submitted material.
 
-   Whistleblower
-	A user that is reporting to the Platform by uploading material and filling out the submission form.
+   Notification
+        A method used to notify a *recipient* of a new *submission*, or an updated existing *submission*.
+	
+   Admin
+        The user that has set up and is maintaining the *platform*.
```
evilaliv3, line 36: Thank you @comradekingu !

I noticed that with your review you have changed the order of the terms.

Was this intended? Maybe we could preserve the alphabetical order?

Thank you!

evilaliv3:
I see your point and i really appreciate your attention to usability here.

For the moment i'm going to integrate the texts revisions postponing for a while the reordering that may result inconsistent with the general organization of the documentation that always uses alphabetical order

PR #3016, opened 2021-05-28
Docs: User Interfaces language reworked

comradekingu:
Went looking for something and found a "Url" in the title.

Depends on: https://github.com/globaleaks/GlobaLeaks/pull/2968

### DIFF
--- a/documentation/user/admin/UserInterface.rst
+++ b/documentation/user/admin/UserInterface.rst
@@ -1,14 +1,14 @@
 User Interface
 ==============
-This section offers you a summary of the user interface offered to Admin users.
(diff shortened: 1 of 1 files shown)


documentation/user/admin/UserInterface.rst
```diff
@@ -161,17 +161,17 @@ Each of the software question types make it possible to configure the following
 
   Question: The text of the question
 
-  Hint: A hint that will be shown via an popover an a question mark near the question.
```
comradekingu, line 164: "an a question mark near the question"
is the part that is neither here nor there.

PR #3043, opened 2021-09-10
Update UserInterface.rst

ujdhesa:
Typo fixing

- [ ] Your code should contain tests relevant for the problem you are solving
- [ ] A description of the problem you're trying to solve
- [ ] If the feature changes current behavior, reasons why your solution is better.


ujdhesa:
There’s also a number of “furtherly” which seem misspelled, but I’d leave that to the opinion of a native…

evilaliv3:
Thank you @ujdhesa !

PR #3148, opened 2022-01-17
Add location to $rootScope so it can activate menu items

fdobrovolny:
I noticed that in the `sidebar.html` templates a call to location is being made in order to activate buttons. However, this fails and the nav links are not activated.


evilaliv3:
Thank you for spotting this regression @fdobrovolny, I'm going to integrate your patch right away!

PR #3280, opened 2022-09-16
Eic dev

cybersmar:


client/app/views/partials/footer.html
```diff
@@ -1,2 +1,2 @@
 <div id="CustomFooter" class="vcenter preformatted mb-1" data-ng-if="public.node.footer" data-markdown-to-html="public.node.footer | stripHtml"></div>
-<div id="AttributionClause" class="vcenter">Powered by <a href="https://www.globaleaks.org/" title="GlobaLeaks - Secure Whistleblowing Software" rel="noreferrer" target="_blank">GlobaLeaks</a></div>
```
evilaliv3, line 2: @cybersmar : Actually this change is not compatible with the GlobaLeaks software license:  https://github.com/globaleaks/GlobaLeaks/blob/devel/LICENSE

You could actually customize the footer but you should not remove the GlobaLeaks AttributionClause

cybersmar: Sorry, my mistake, it won't happen again.

PR #3308, opened 2022-11-04
date range picker implemented using directives

dev-genesistechnologies:


evilaliv3:
Very good work @dev-genesistechnologies 

I noticed very little defects and i consider this entire work is ready for integration

On improvement that i would consider adding is to give the color #377ABC to every icon when the filter is active.
This would help users identify which are the activated filters. Without this i see in fact that one can activate multiple filter and see no data and forget which filter they activated.

Do you mind adding this last correction?

dev-genesistechnologies:
> 

sure will go through it in the morning

dev-genesistechnologies:
> Very good work @dev-genesistechnologies
> 
> I noticed very little defects and i consider this entire work is ready for integration
> 
> On improvement that i would consider adding is to give the color #377ABC to every icon when the filter is active. This would help users identify which are the activated filters. Without this i see in fact that one can activate multiple filter and see no data and forget which filter they activated.
> 
> Do you mind adding this last correction?

@evilaliv3

evilaliv3:
Thank you @dev-genesistechnologies . Well done!

I think we could proceed integrating your work.

I will proceed mergin your pull inside the branch feature/datafilters; then i will apply some cleanhup of the branch to help who would eventually need to audit the feature as a whole and then merge it in devel preserving your credits.

client/app/js/directives.js
```diff
@@ -59,8 +63,18 @@ GL
...
+         if ($scope.daterangePickerModel.start && $scope.daterangePickerModel.end) {
+           $scope.obj.value1 = true;
+         }else{
+           $scope.obj.value1 = false;
+         }
+
          $scope.onDatechange($scope.dataRangeFilter);
       });
+      $scope.checkFilter = function() {
+        return $scope.obj.value1 === true;
```
evilaliv3: @dev-genesistechnologies : i will simply remove this variable and change this function to do:

return "return $scope.daterangePickerModel.start && $scope.daterangePickerModel.end;"

Issue #3348, opened 2023-02-13
Make it possible to visualize Images files on the platform


pvigoseidor:
As a recipient user I want to visualize the images files on GlobaLeaks without download the attachment.

It would be fantastic the possibility to view directly the attachment image files on platform.

PR #3377, opened 2023-03-02
Feature/file viewer

pnellikkaparambil:
Added support for visualizing files on the platform

- Added support for Image #3348

### DIFF
--- a/client/app/views/partials/tip_files_receiver.html
+++ b/client/app/views/partials/tip_files_receiver.html
@@ -12,7 +12,7 @@
         <thead>
           <tr>
             <th data-translate>Filename</th>
-            <th data-translate>Download</th>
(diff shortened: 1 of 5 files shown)


client/app/views/partials/tip_files_receiver.html
```diff
@@ -12,7 +12,7 @@
         <thead>
           <tr>
             <th data-translate>Filename</th>
-            <th data-translate>Download</th>
+            <th data-translate>Download / View</th>
```
evilaliv3, line 15: This should be split like:

```html
<th><span data-translate>Download</span> / <span data-translate>View</span></th>```

evilaliv3 (changes requested):

PR #3561, opened 2023-07-25
audio masking implemented on top of audio recording

msmannan00:
1. low pass filter applied for reducing unwanted frequencies
2. pitch shift applied to modify pitch and duration of audio
3. noise reduction optimized to reduce or remove background noise considering gossip cancelation
4. audio file size optimized
5. overall code refactored

### DIFF
--- a/client/app/js/services.js
+++ b/client/app/js/services.js
@@ -42,9 +42,9 @@ factory("Authentication",
       };
 
       self.reset = function() {
-	self.loginInProgress = false;
-	self.requireAuthCode = false;
-	self.loginData = {};
+  self.loginInProgress = false;
+  self.requireAuthCode = false;
+  self.loginData = {};
       };
 
       self.login = function(tid, username, password, authcode, authtoken) {
@@ -80,7 +80,7 @@ factory("Authentication",
 
         };
 
-	var failure_fn = function(response) {
+  var failure_fn = function(response) {
           self.loginInProgress = false;
 
           if (response.data && response.data.error_code) {
@@ -1239,7 +1239,7 @@ factory("Utils", ["$rootScope", "$http", "$q", "$location", "$filter", "$timeout
             deferred.resolve(response);
           },
           function() {}
-	);
+  );
       }
 
       return deferred.promise;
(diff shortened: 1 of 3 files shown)


evilaliv3:
Thank you @msmannan00 

The patch seems really good.
Few points to address:
1.
Pitch switch
What we need to change is discussed is the pitch switch where it is better to avoid to have a random pitch and have always the same expact pitch switch moving the recording to [Bass tone (E2 to E4 voice type)](https://en.wikipedia.org/wiki/Voice_type)
Is it possibly for you to apply this change?

2.
Gossip removal: would you please privide reference documentations for this modification and why it matters?

3.
Noise suppression: i remember you evaluated to use [web audio noise suppression](https://developer.mozilla.org/en-US/docs/Web/API/MediaTrackSettings/noiseSuppression); is this still the case? if now would you please clarify the reasons why you consider it is not good to use it?

thank you!

client/app/js/services.js
```diff
@@ -1277,6 +1283,108 @@ factory("Utils", ["$rootScope", "$http", "$q", "$location", "$filter", "$timeout
...
+    applyGossipRemoval: function (buffer) {
+      const blockSize = 128; // Adjust the block size as needed (a power of 2 is recommended)
+      const numBlocks = Math.ceil(buffer.length / blockSize);
+
+      for (let blockIndex = 0; blockIndex < numBlocks; blockIndex++) {
+        const blockStart = blockIndex * blockSize;
+        const blockEnd = Math.min(blockStart + blockSize, buffer.length);
+
+        for (let i = blockStart; i < blockEnd; i++) {
+          if (Math.abs(buffer[i]) < 0.03) {
```
evilaliv3, line 1325: @msmannan00 : which is the rationale behing this function and is there any documentation reference for this comparison? why exactly 0.03?

msmannan00:
So basically noise suppression and gossip removal are same thing only the function name has changed. I have been testing it for some hours by taking mobile and playing some talking sounds with different volumes at different locations in room and adjusting the noise removal so that the conversation in background is removed. This was something minimized in low pass filter but not completely removed

evilaliv3:
I see thank you for clarifying.

As previously discussed what i think we really miss is to document this aspects properly.
If this functions are standard functions, we could add them to the code and document their use in our documentation referring to literature. If there is no literature unfortunately we wont be able to use them.

I reformulate what i propose trying to make it more clearer what we need to do now:
1) pitch switch: we need a standard function widely used that we can implement AS IS. The function ideally should enable to define the tone as frequency so should by a function like pitchSwitch(audiofile, frequency)
2) noiseSuppressor: i think at the moment if we miss a standard function we shoiud just rely on standard noisesuppression property of the WEB audio API https://developer.mozilla.org/en-US/docs/Web/API/MediaTrackSettings/noiseSuppression. Is basically just a boolean that we need to set on the track that will activate noisesuppressors on microphones

evilaliv3:
Closing as the code has been merged manually.

Code for pitch change currently not integrated waiting for some additional research on the matter.

PR #3639, opened 2023-09-19
statistics fixes and improvement to support latest devel branch

msmannan00:


msmannan00:
old statistics pull request had conflicts and got corrupted due to different backend versioning. support for latest devel implemented

PR #3651, opened 2023-09-22
SEO improvements

gingerlime:
* see https://github.com/orgs/globaleaks/discussions/3633
* Google Lighthouse does not like `href=""` but `href="#"` is fine


evilaliv3:
Thank you @gingerlime

PR #3675, opened 2023-09-29
Feature/3588 pdf export

pnellikkaparambil:
Update print layout with
* Removed header, footer and tooltip


evilaliv3:
Thank you!

I've integrated the patchs manually preserving your autorship.

PR #3718, opened 2023-10-24
add button export csv on reports list

mttlinks:


evilaliv3:
Thank you @mttlinks 

I'm going to remove the commit 20e2dcedef264b5d27d33b7996bb5b63b98b0b7a from this pull request since the pull request should include only the specific commits of the feature implemented.

Issue #3848, opened 2023-12-08
Typo in docker user


dahaupt:
### What version of GlobaLeaks are you using?

v4.13.18

### What operating system(s) are you seeing the problem on?

Linux

### Describe the issue

Error when starting the docker container:
```
Error response from daemon: unable to find user globleaks: no matching entries in passwd file
```

PR #3849, opened 2023-12-08
Fix typo in docker user

dahaupt:
Resolves: #3848


evilaliv3:
Thank you @dahaupt. 

I've reviewed your patch and integrated it manuall since there was one conflict preserving your contribution authorship!

PR #3887, opened 2023-12-14
added helm chart for globaleaks usage for kubernetes

Andrew-Moy:
Updated pull request for helm chart for kubernetes usage of globaleaks. This has been rebased in accordance to the newest app version of globaleaks.


rglauco:
@Andrew-Moy readinessProbe and livenessProbe are performed towards values.service.port, which is 80. In my cluster i obtain error, since the exposed containers ports are 8080 and 8443. In my helm deployment i use that values https://github.com/rglauco/globaleaks-helm/blob/main/globaleaks/values.yaml

Andrew-Moy:
@rglauco Thank you, I have updated the ports in my values.yaml

evilaliv3:
@Andrew-Moy @rglauco : At the moment on the repositoy we have a directory /docker that contains the docker scripts, and now we are adding a directory /charts.

If you are not aware of more standard alternative, what about creating a directory /containers and move the directory /docker and /charts inside that? what do you think?

Andrew-Moy:
> @Andrew-Moy @rglauco : At the moment on the repositoy we have a directory /docker that contains the docker scripts, and now we are adding a directory /charts.
> 
> If you are not aware of more standard alternative, what about creating a directory /containers and move the directory /docker and /charts inside that? what do you think?

@evilaliv3 I don't mind that idea, however I think that keeping charts and any potential future charts separate from the docker scripts is fine for clarity. Will ultimately go with what you think is better

evilaliv3:
Sure, I also think they should be separated in subdirectories within /containers so that we could put here different subdirectories related to different containers technologies.

Andrew-Moy:
Got it, will implement this change along with some more bug fixes in the upcoming days

PR #3982, opened 2024-01-30
client update to angular 16

msmannan00:
1. Angular updated to angular 16
2. libraries updated to support latest angular
3. older client moved to legacy folder


.gitignore
```diff
@@ -0,0 +1,66 @@
+backend/env
```
evilaliv3, line 1: @msmannan00 would you please remove this file and squash the commit?

msmannan00:
> @msmannan00 : For the moment i suggest to replace this patch and just have a console.log() where we use `document.querySelector` to log something simple like "Selector not found: $selectorname".
> 
> The reason is that we don't know which could be the effect of unloading the CSS/Javascript of the user; the triggered code will have anyhow already applied changes to the page that we could not unload.
> 
> What do you think?

its just a suggestion to implement route reloading without refreshing the page if a JavaScript error occurs, and avoid loading the JavaScript this time. This approach would ensure that nothing gets messed up. Otherwise, if we don't adopt this strategy, numerous issues could arise when we release the code. Users might attempt to use the code despite the errors, leading to undesirable outcomes

evilaliv3:
> its just a suggestion to implement route reloading without refreshing the page if a JavaScript error occurs, and avoid loading the JavaScript this time. This approach would ensure that nothing gets messed up. Otherwise, if we don't adopt this strategy, numerous issues could arise when we release the code. Users might attempt to use the code despite the errors, leading to undesirable outcomes


i see.
thank you for your feedback @msmannan00 

At least for the initial release i suggest anyhow to avoid to include this patch and think to it a little further.
Lets add eventually just a log about the found selectors.

I really think in fact that:
- removing a loaded javascript will not rever the changes that the javascript will have already applied.
- removing a javascript at the first error will prevent users to possiblly identify which are the parts that do not function; with a log on the not-found selectors we could collect feedback to fix the custmization

msmannan00:
> > its just a suggestion to implement route reloading without refreshing the page if a JavaScript error occurs, and avoid loading the JavaScript this time. This approach would ensure that nothing gets messed up. Otherwise, if we don't adopt this strategy, numerous issues could arise when we release the code. Users might attempt to use the code despite the errors, leading to undesirable outcomes
> 
> 
> i see.
> thank you for your feedback @msmannan00 
> 
> At least for the initial release i suggest anyhow to avoid to include this patch and think to it a little further.
> Lets add eventually just a log about the found selectors.
> 
> I really think in fact that:
> - removing a loaded javascript will not rever the changes that the javascript will have already applied.
> - removing a javascript at the first error will prevent users to possiblly identify which are the parts that do not function; with a log on the not-found selectors we could collect feedback to fix the custmization

Agreed, for this release will make changes to remove the revert functionality while keeping selector logs, user would be able to see error logs and make changes accordingly for their scripts. As discussed earlier we can also mention it in application change log for globaleaks 5

evilaliv3:
Closing since all the code is now integrated in branch main and released in version 5.0.0

PR #4009, opened 2024-02-22
Correct error handling for duplicate user error

### LINKED ISSUE #3875: Error creating a user with same username

4.13.21

### What browser(s) are you seeing the problem on?

Chrome

### What operating system(s) are you seeing the problem on?

Windows, Linux

### Describe the issue

It reaches an INTERNAL SERVER ERROR when trying to create a user with an existing username.

Screenshot from 2023-12-13 08-04-36

### Proposed solution

It should check if the username exists before trying to create the new user with the specified username and notify the admin that already exists

jmcpo15:
If sqlalchemy returns a specific error which indicates the user id already exists, then preform a rollback as the try/except clause stops this from happening before returning a more detailed error message.

Added this error into the list of errors in the client.

Resolves #3875 

- [ ] ~~If the pull requests changes current behavior, reasons why your solution is better.~~

### DIFF
--- a/backend/globaleaks/handlers/admin/user.py
+++ b/backend/globaleaks/handlers/admin/user.py
@@ -62,6 +62,10 @@ def db_create_user(session, tid, user_session, request, language):
     if not request['username']:
         user.username = user.id = uuid4()
 
+    existing_user = session.query(models.User).filter(models.User.tid == user.tid).first()
+    if existing_user:
+        raise errors.DuplicateUserError
+
     user.salt = GCE.generate_salt()
 
     user.language = request['language']
--- a/client/app/views/partials/messageconsole.html
+++ b/client/app/views/partials/messageconsole.html
@@ -31,6 +31,10 @@
       <div data-ng-switch-when="14">
         <span data-translate>The upload request exceeds the size limit</span>
       </div>
+
+      <div data-ng-switch-when="17">
+        <span data-translate>A user with this username already exists</span>
+      </div>
     </div>
   </div>
 </div>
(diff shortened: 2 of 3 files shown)


client/app/views/partials/messageconsole.html
```diff
@@ -25,12 +32,18 @@
...
+        >
       </div>
 
       <div data-ng-switch-when="14">
         <span data-translate>The upload request exceeds the size limit</span>
       </div>
+
+      <div data-ng-switch-when="17">
+        <span data-translate>A user with this username already exists</span>
+      </div>
```
evilaliv3, line 46: Please revise the changes to this file including only the addition of error file 17 and exclude every refactor when proposing a specific patch.

backend/globaleaks/handlers/admin/user.py
```diff
@@ -62,6 +62,10 @@ def db_create_user(session, tid, user_session, request, language):
     if not request['username']:
         user.username = user.id = uuid4()
 
+    existing_user = session.query(models.User).filter(models.User.username == user.username).first()
+    if existing_user:
```
evilaliv3, line 66: Please add models.User.tid == tid to ensure to apply this check per tenant, or the check will prevent different tenants' users to use the same username.

evilaliv3 (commented):
Thank you @jmcpo15 

I reviewed your patch that is great. I left you some comments about the changes to be applied before the merge.

Please let me know if you would like to author them and please then squash your commits into one.

jmcpo15:
@evilaliv3 All of those changes have been made and squashed into one commit

evilaliv3:
@jmcpo15: Actually now you are looking for the first user given a tenant.

Please edit your code to include ```user.tid == tid, user.username == username```

evilaliv3:
Pull request merged and i'm going to apply the correction on the on top of it. thank you!

PR #4030, opened 2024-03-21
fix: css tos-text expansion to full page width

rglauco:

- [ x ] The pull request should include a description of the problem you're trying to solve
- [ x ] The pull request should include overview of the suggested solution
- [ x ] If the pull requests changes current behavior, reasons why your solution is better.
- [ x ] The proposed code should be fully functional
- [ x ] The proposed code should contain tests relevant to prove is functionality
- [ x ] The proposed tests should ensure significative code coverage
- [ x ] All new and existing tests should pass

### DIFF
--- a/client/app/css/main.css
+++ b/client/app/css/main.css
@@ -847,7 +847,7 @@ text.axis-worktime {
   font-family: monospace;
   font-size: 0.8rem;
   max-height: 40em;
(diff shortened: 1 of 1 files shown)


evilaliv3:
Thank you @rglauco !

Issue #4025, opened 2024-03-19
Contact details of Whistleblower can not be masked


elbill:
### What version of GlobaLeaks are you using?

4.14.8

### What browser(s) are you seeing the problem on?

All

### What operating system(s) are you seeing the problem on?

Windows

### Describe the issue

Can not mask whistleblower contact details when template override exists.
Have not tried it without template override.

### Proposed solution

Masking should work for whistleblowing details with and without template override.

evilaliv3:
Thank you for reporting this @elbill 

@msmannan00: would you please give it a look performing first a patch for GlobaLeaks 4?

evilaliv3:
Assigning this to me to verify if this is actually still the case.

evilaliv3:
i've verified this an i can confirm that all was perfectly solved back in time.

PR #4035, opened 2024-03-25
masking feature implemented on user identity

msmannan00:
This pull request addresses issue #4025, which reported that whistleblower contact details could not be masked when a template override existed. The feature has now been implemented to ensure that whistleblower identity is properly protected, even in cases where a template override is in place.

Changes:
Added functionality to mask whistleblower contact details with template overrides.


evilaliv3:
Thank you! merged manually

PR #4047, opened 2024-04-17
[WIP] Add Dockerfile which builds globaleaks from source

kskarthik:

### DIFF
--- a/docker/Dockerfile.source
+++ b/docker/Dockerfile.source
@@ -1,0 +1,38 @@
+# Author: Sai Karthik <kskarthik@disroot.org>
+# Summary: This dockerfile builds globaleaks from source
+# The build is done in multistage
+
+# Build the webapp
+FROM node:lts-alpine as webapp
+
+COPY . /globaleaks/
+
+WORKDIR /globaleaks/client
+
+RUN echo "installing dependencies" && \
+		npm i --incude=dev && \
+		echo "building webapp ..." && \
+		./node_modules/grunt/bin/grunt build && \
+		ls -al
+
+# Build the backend
+FROM python:3.12-slim
+
+RUN useradd -m -U gl
+
+USER gl
+
+# copy the globaleaks dir form the previous stage
+COPY --from=webapp /globaleaks /home/gl/globaleaks-src
+
+WORKDIR /home/gl/globaleaks-src/backend
+
+RUN echo "creating workingdir" && \
+		mkdir ~/globaleaks-data && \
+		echo "installing backend dependencies" && \
+		pip install -r requirements.txt && \
+		ls -al
+
+EXPOSE 8080 8443
+
+CMD ["python3", "bin/globaleaks", "-n", "--working-path=/home/gl/globaleaks-data"]


kskarthik:
@evilaliv3 need your suggestions. Please review

evilaliv3:
Thank you @kskarthik for your contributions.

Actually i've some important comment. Feel free to provide your feedback.

At globaleaks we try to be as reproducible as possible and to test every aspect of the setups.
1. Current builds are made for debian and the software is fully tested and optimized for debian that features reproducible builds in their development lifecycle: https://en.wikipedia.org/wiki/Reproducible_builds
2. GlobaLeaks is built using python python dependencies and during development we use PIP to ease the setup of the dependencies, but for production we build a debian package installing those dependencies and other dependencies (a firewall, apparmor, and other tools) as debian package.
3. It is for reason 1 and 2 that the current docker image is build using debian bookworm as base image and inside that we do not build the globaleaks package but we follow the official globaleaks installation guide running the script that we wrote that add our repository to the operating system and install the official debian package that we release.

For this reason if you are interested supporting optimizing and automating the setup this is more than welcome but we should try to preseve the above conditions.

kskarthik:
Ok, Great! I got the picture. Thank you for the detailed explanation.

So, The ideal way is to use the existing docker file in the github actions which will run when a new tag is pushed.

## I can imagine the release flow briefly as:  

Building -> Testing -> debian packaging  -> pushing to debian repo -> docker image from the debian package

Feel free to add more thoughts!

evilaliv3:
Yes this is doable.

Right know we tag (that is actually a signature) before the release and use the tag build the package.

But i think that:
- we could start releasing using the commit has of the relase commit (e.g. https://github.com/globaleaks/GlobaLeaks/commit/765b5e515763c9472a4e8d52ce1978da6ce443ad)
- we could use the tag to trigger the automated releases and eventually for the future find a way to have subsequent actions that first build and release the package and then proceed publishing the docker image)

I'm actualy not super sure that docker enable to publish via github in automated fashion when using a free account. have you already verified this possibility?

kskarthik:
> I'm actualy not super sure that docker enable to publish via github in automated fashion when using a free account. have you already verified this possibility?

Yes, we can! I do it for my repo & mine is a free account https://github.com/kskarthik/indian-fincodes-api/blob/main/.github/workflows/docker.yml

kskarthik:
@evilaliv3 i found a use case for this PR. 

We could build this docker image with a `nightly` or `dev` tag for each commit & upload it to the docker repo. Users who wish to checkout what's new in globaleaks can use this image. they do not have to setup build env. And this is norm with most FOSS projects which provide a nightly build for faster feedback loops etc...

We have to explicitly mention in docs that this docker image must be used for evaluation only.

Do you align with this idea? I will add few more updates to the PR like adding a github action which builds the image.

evilaliv3:
Thank you @kskarthik 

Actually eventually having a nightly build of the devel branch is interesting but still i consider that it should eventually continue to try to replicate a production setup, using Debian as base image and the official build scripts that we use for official release.

kskarthik:
i would propose production way for stable releases & docker way for nighly one's

docker images are easy to install & dispose than debian counterpart for evaluation purposes.

PR #4054, opened 2024-04-22
improve the documentation

kskarthik:
* Setup github actions to build the site which builds the docs when there is a change made
* Add info about the dev setup for building the docs site


evilaliv3:
Thank you @kskarthik 

Would you please clarify why you consider this changes to be helpful?
GlobaLeaks try to keep the codebase and the functionalities small and reduce over-engineering that is key for ensuring the project could be governed. We suggest as well that  changes and development follows first and evaluation and a proposal with a ticket, to reduce unneeded work and plan and track together all the changes.

Few notes:
- At the moment the documentation is already continuously published on doc.globaleaks.org at every single commit
- The pull request contains changes to the content of the documentation which should be eventually proposed separately to ensure every change is evaluated properly by topic.

kskarthik:
> Would you please clarify why you consider this changes to be helpful?

I just added a simple job to make sure that every MR or push which changes the content inside the`documentation/` folder does not break the docs site in prod. I did not consider it as a over-engineering, but a safely check.

> We suggest as well that changes and development follows first and evaluation and a proposal with a ticket, to reduce unneeded work and plan and track together all the changes.

Yes, i should have proposed this idea first, before sending the patch.

> At the moment the documentation is already continuously published on doc.globaleaks.org at every single commit

Then, I wonder why 59de60466894db06014a568934eb284d91eef27d bd34480faca26d754ab5e47886be9ef5997e9179 were not reflected in the docs, as of writing this comment

kskarthik:
> The pull request contains changes to the content of the documentation which should be eventually proposed separately to ensure every change is evaluated properly by topic.

Sure

kskarthik:
Now this PR contains only the CI job patch

PR #4064, opened 2024-04-30
Add Version bump script

kskarthik:
This script will automate the chore of changing of versions across multiple files when there is a release.

It expects to be run from the project's root:

Example:  `GlobaLeaks/scripts/bump_version.sh 4.30.3 `

The `CHANGELOG` file only must be manually edited.

This saves the valuable developer's time.

### DIFF
--- a/scripts/bump_version.sh
+++ b/scripts/bump_version.sh
@@ -1,0 +1,32 @@
+#!/bin/bash
+# This script bumps the project version across multiple files
+# The script is expected to be run from the project's root directory
+# with new version as the first argument
+set -e
+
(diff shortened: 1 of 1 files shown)


kskarthik:
please let me know if anything changed, I will cover that file as well

kskarthik:
@evilaliv3 done!

evilaliv3:
@kskarthik : merged with few changes.

I've removed the edit to the CHANGELOG file that from now on we could author in advance, before calling the script.

Instead of the edit to the CHANGELOG i preferred to add:
- the commit with message "Bump to version $version"
- the creation of the release tag

Issue #4065, opened 2024-05-01
Automating the Publishing of docker image


kskarthik:
### Proposal

I have a thought regarding automating updating of docker image to the project's docker registry.

The workflow is described in-depth below:

## Tasks to be done by the developer locally:

1. creates a tag
2.  build the deb locally & sign
3. push to the project's debian archive
4. push the tag to github

## The Proposed Automation part:

5.  A workflow will be tiggered on github actions when a tag is pushed
6. A script checks if a new version which matches the git tag, is published at [deb.globaleaks.org](http://deb.globaleaks.org/)
7. If yes, It builds docker image from the Dockerfile present in the project's repo & tags it with the latest version. Else, the job fails.
8. Pushes to the docker hub

### Motivation and context

This will reduce the number of repetitive manual tasks to be done for each release & helps the developers prioritize their attention

kskarthik:
i have a proof-of-concept implementation for this method.

evilaliv3:
Closing as addressed since we are not able to build and publish images automatically on dockerhub.

Thank you @kskarthik !

PR #4066, opened 2024-05-03
Automate docker image build on each release

kskarthik:
A docker image will be built for each release
Only builds if there is a matching release in the debian repository

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,23 @@
+name: Release Globaleaks
+
+on:
+  push:
+    # this workflow only runs on tag names prefixed with "v"
+    # and suffixed with "-docker"
+    tags:
+      - "v*-docker"
+
+jobs:
+  publish_docker_image:
+    runs-on: "ubuntu-latest"
+    steps:
+      - name: Check out repository code
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Build
+        env:
+          DOCKER_HUB_TOKEN: ${{ secrets.DOCKER_HUB_TOKEN }}
+        run: |
+          ./.github/workflows/scripts/build_docker.sh
--- a/.github/workflows/scripts/build_docker.sh
+++ b/.github/workflows/scripts/build_docker.sh
@@ -1,0 +1,42 @@
+#!/bin/bash
+set -e
+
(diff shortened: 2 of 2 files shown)


kskarthik:
The PoC implementation of proposal #4065

kskarthik:
Please note that `DOCKER_HUB_TOKEN` secret must be set in the repo

kskarthik:
@evilaliv3 any change my PR's get reviewed now ? Since v5.0.0 is now released as mentioned in the past

evilaliv3:
> @evilaliv3 any chance that my PR's get reviewed now ? Since v5.0.0 is now released as mentioned in the past

Definitely, thank you @kskarthik 

Actually not we have a docker sponsored acount so that i should be able to have a token to automate builds.

I left you a comment, do you think such a change would be possible?

kskarthik:
> I left you a comment, do you think such a change would be possible?

I did not get the context, the script accepts the `DOCKER_HUB_TOKEN` env which you have to set in the github secrets

evilaliv3:
@kskarthik: i'm referring to the fact to trigger the build only when we issue a tag  like "v5.0.0-docker" starting with "v" and ending with "-docker".

kskarthik:
@evilaliv3 understood & implemented the same!

kskarthik:
> I left you a comment

I guess you were referring to slack message, which i no longer can access

evilaliv3:
Thank you @kskarthik 

Actually i found out that with the sponsored github account we can use the auto-build function.

Ive tried to add a rule on dockerhub that anytime we generate a tag with the following regexp generated a build: ```^v[0-9]+.*-docker$```

This simplify the configuration not requiring to implement a gitaction and not requesting to manually configure secrets.

this is very cool since now the logs of the build are preserved directly on dockerhub along with the build

kskarthik:
Ok, cool

evilaliv3:
Thank you anyway @kskarthik! 

Its an advantage to finally have a sponsored dockerhub account, but your patch was cool!

kskarthik:
> Its an advantage to finally have a sponsored dockerhub account

True, It avoided a hassle of code maintenance.


Please also review my other PR's: https://github.com/globaleaks/GlobaLeaks/pulls/kskarthik

Issue #4069, opened 2024-05-08
Turning off notifications to account after first report


schris-dk:
### Proposal

Create a default setting for an account to ONLY receive notification on NEW reports (and not changes to existing ones)

### Motivation and context

We have clients, who uses a different entity (KL) to recieve and validate incoming reports. KL is handling a large amount of clients, hence, they have many sýstems to handle. For each system, they have personal accounts as well as a common group account, to monitor incoming reports from all systems (to ensure, that no reports are 'drowned' in the personal account mailboxes). This account is not used to do any case handling.  

We are aware of, that notifications can be disabled on each case seperately, but this requires login to each case to set this value, so to easen the burden, a global setting for the account could come in handy. KL informed us, that the other systems, that they used has this feature

evilaliv3:
Thank you for this proposal @schris-dk 

Do you have any idea where to add this option an how to revise the UI and text to add it?

Are you able to provide a mockup of the revised interface?

schris-dk:
I'd suggest this setting in the user profile

![image](https://github.com/globaleaks/GlobaLeaks/assets/70965554/bfacf92a-3e89-4478-aead-e5cdb11cf00a)

PR #4086, opened 2024-05-29
Turning off notifications to account after first report

yzoug:

# Description

This is still a draft, work in progress. First PR here :)

Implements a solution to issue #4069. A new field, `tip_update_notification`, is added to the User model and a migration adds this field to existing Users. The UI is updated to allow modification of this preference, and the notification logic is updated to only notify a recipient of a tip update if this setting is at True (the default value).

This is not fully functional yet, I have a "ProgrammingError" SQLite exception when running this, and I did not yet fully test it. Will keep working on this when time allows, I am of course open to any help/suggestions from the maintainers.

### DIFF
--- a/backend/globaleaks/__init__.py
+++ b/backend/globaleaks/__init__.py
@@ -10,7 +10,7 @@ __copyright__ = '2011-2024 - GlobaLeaks'
 __version__ = '4.15.4'
 __license__ = 'AGPL-3.0'
 
-DATABASE_VERSION = 68
+DATABASE_VERSION = 69
 FIRST_DATABASE_VERSION_SUPPORTED = 52
 
 # Add new languages as they are supported here! To do this retrieve the name of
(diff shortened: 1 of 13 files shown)


evilaliv3:
Thank you @yzoug 

The patch as idea is interesting and the patch very accurate and minimal, that is great and  there are as well very few defects like for example it includes some changes not related to the patch. It also seems to me to not include any defects except some changes to the file fr.po that should not be included.

As an exercise of patch it is perfect, but what i'm not sure is about the feature offered as it has some drawbacks: if we do not notify updates anyhow users will start receiving reminders about unread reports. One can think to silence also these but this then will lead to users to miss everything; Maybe we should instead implement a feature that say specifically: "Notify only reports creation" or something like this. I consider that should probably discuss more about the possibilities.

yzoug:
Thanks @evilaliv3! You're right, I didn't think about the unread report notification emails, that's a problem with the current implem for sure. I'll close this PR for now, but I'll try to come to the Slack to discuss this, or any other good first issue that I can help out with.

evilaliv3:
You are welcome! thank you!

PR #4120, opened 2024-06-26
Feature/tenant profiles

pnellikkaparambil:
This pull request relates to a first attempt of implementation of "Tenant" (Sites) profiles to make it possible to create preconfigurations to be used as defaults for new sites.

It extends the Tenant table by adding an attribute "is_profile".
When the Tenant has the profile attribute set as "true" the site is considered to be a site profile.

The patch exports this variable "is_profile" to the backed API and use it on the client in order to be able to show separatel actual tenant instances (is_profile=False) and tenant profile (is_profile=True)

The patch includes as well a partial implementation of the migration script necessary for porting an old database schema to a new database schema including the new attribute Tenant.is_profile.


evilaliv3:
Closing since the code is now integrated in branch feature/profiles

PR #4151, opened 2024-08-09
tenant optimization

msmannan00:
1. Integrated profiles.
2. Created a default profile for tenant 1000000, applied in both config and config_lang.
3. Implemented migration scripts.
4. Removed duplicate variables from config and config_lang, mapping them to the default profile.
5. Implemented chaining for tenants to first search for variables in their profile, and if not found, default to the default profile.


evilaliv3:
Closed since the code is not integrated in branch feature/profiles

PR #4176, opened 2024-09-02
Update publiccode.yml

gionni87:
Inserita l'agenzia ANSFISA, URL pubblico https://whistleblowing.ansfisa.gov.it


evilaliv3:
Thank you @gionni87 

Actually at the moment, for reasons of simplicity, we are just listing:
- a very limited set of ministries related to whistleblowing (the first orgs that you see listed)
- the users of whistleblowing.it for which we maintain a verified list: https://www.whistleblowing.it/adesioni/

It is in fact very difficult for us to follow and verify all the other organizations over time.

For this reason i dont think we could integrated you, since we have no way to track integrations like this.

I hope you understand.

PR #4277, opened 2024-10-21
Update codebase to use new angular 17 standalone components and control flow

msmannan00:
1. Replaced deprecated control flow directives with the latest syntax.
2. Converted components to standalone, removing NgModules.
3. Updated route definitions with standalone component-based routing.
4. Optimized lazy loading using new route-based syntax.
5. Cleaned up dependencies and ensured compatibility with the latest Angular version.


evilaliv3:
@msmannan00 : This is an example of "regression" where with your proposal you are reintroducing some old code.

I will give it a try to find other situations like this.

image

msmannan00:
Yup, actually in past two weeks I tried to take merge from stable version twice and make changes accordingly but even then I might have missed some commits that were made in the week

evilaliv3:
Very good to go! Merging!

PR #4287, opened 2024-10-31
Feature/profiles

msmannan00:
feature profile and config optimization


evilaliv3:
@msmannan00: i performed a first run of tests and all seems in very perfect shape.

I took care of merging the the devel branch into this and solving the conflicts:

At the moment i notice the following just the following defects that we could ameliorate
- config.'version', 'last_version', 'db_version' appear at the moment present in both tenant 1 and profile 1000001; i wonder if to keep this or to find ways of having these only either in tenant 1 or profile 100001. The same could be said for example in relation to variables like profile_counter that is present both in tenant1 and profile 1000001 but actually this second variable appears not used.
- config.onion and config.tor_onion_key appear at the moment present also on profiles: maybe these variables should be present only on tenants? i see in fact that at the moment we are also instantiating an onion service for the tenants profiles.

I will continue looking into this next week.

msmannan00:
Yes, you are right about that, it should only be on tenant I guess I missed it from my side

evilaliv3:
@msmannan00 , apparently this tests seems to fail once every two runs: globaleaks.tests.handlers.recipient.test_rtip.TestRTipInstance.test_delete

```
globaleaks.tests.handlers.recipient.test_rtip
  TestRTipInstance
    test_delete ...                                                     [ERROR]

===============================================================================
[ERROR]
Traceback (most recent call last):
  File "/home/evilaliv3/Development/whistleblowing-software/backend/env/lib/python3.12/site-packages/twisted/internet/defer.py", line 1656, in _inlineCallbacks
    result = current_context.run(
  File "/home/evilaliv3/Development/whistleblowing-software/backend/env/lib/python3.12/site-packages/twisted/python/failure.py", line 514, in throwExceptionIntoGenerator
    return g.throw(self.type, self.value, self.tb)
  File "/home/evilaliv3/Development/whistleblowing-software/backend/globaleaks/tests/handlers/recipient/test_rtip.py", line 289, in test_delete
    yield handler.delete(rtip_descs[0]['id'])
  File "/home/evilaliv3/Development/whistleblowing-software/backend/globaleaks/tests/helpers.py", line 128, in callInThreadWithCallback
    result = func(*args, **kw)
  File "/home/evilaliv3/Development/whistleblowing-software/backend/globaleaks/orm.py", line 185, in _wrap
    result = function(session, *args, **kwargs)
  File "/home/evilaliv3/Development/whistleblowing-software/backend/globaleaks/handlers/recipient/rtip.py", line 795, in delete_rtip
    raise errors.ForbiddenOperation
globaleaks.rest.errors.ForbiddenOperation: ForbiddenOperation: <<Operation Forbidden>> (8) HTTP:403

globaleaks.tests.handlers.recipient.test_rtip.TestRTipInstance.test_delete
```

The failure seems related to the user profile permission: ```profile.can_delete_submission```

msmannan00:
let me check it out, the issue was there but fixed it last week, most probably was skipped in commits will review it out asap

msmannan00:
@evilaliv3 test is fixed, most probably it wont fail again, good to see branch is already on latest codebase

evilaliv3:
@msmannan00: following our community meeting i've have some considerations that i consider we would probably need to solve.

1) shall we make available to a tenant that uses a tenant profile the user profiles defined in that tenant profile?
2) if yes, what shall we do if a tenant is reconfigured to use a different tenant profile?
3) shall we make available to a tenant that uses a tenant profile the questionnaires defined in that tenant profile?
3) if yes, what shall we do if a tenant is reconfigured to use a different tenant profile?
4) there are some systems with several tenants (e.g. 5000) each with one single user configured with the same configuration there maybe administrators will be very frustrated of viewing 5000 profiles. Probably we need to find a strategy to avoiding showing visually the profile existence for the profiles that we create on migration and just configure special profiles and reuse the angular partial to be shown below the user configuration. I will clarify about this point in our next call as it will be more easy talking about this.

PR #4339, opened 2024-11-29
fix/multistep-questionnaire

msmannan00:
fix(multistep-questionnaire): bind required file/audio upload check to model

Fixed a bug in the multistep questionnaire where the required check for file or audio uploads was not functioning correctly. Updated the binding of the "required" condition to use the actual model fields, ensuring proper validation.

### Problem
The validation for required file/audio uploads in the multistep questionnaire was not functioning as intended because the condition was not bound to the actual model fields and caused required check to skip

### Solution
The "required" condition was updated to bind directly to the model fields, ensuring that the validation logic works as expected. logic for required field was also fixed and updated


evilaliv3:
Fantastic! I retested it and i've not found any defect.

Merging it!

PR #4367, opened 2024-12-30
Replace SecureTemporaryFiles with an EphemeralFile implementation based on ChaCha20

evilaliv3:
With this pull request i would like to propose:
- Replacement of SecureTemporaryFiles based on AES with EphemeralFile based on ChaCha20
- Addition of the globaleaks-eph-fs command that we might use to instantiate an ephemeral filesystem when integrating ClamAV and MAT2


evilaliv3:
@msmannan00 : When you like please feel free to proceed with the review of this pull request.

evilaliv3:
Actually i missed to add the libraries! Going to fix it fist before opening for review.

Thank you!

msmannan00:
Yes will do so thanks

PR #4370, opened 2025-01-02
Refactor SecureTemporaryFile using ChaCha20 in place of AES

evilaliv3:
With this pull request i would like to propose the refactoring of SecureTemporaryFile using ChaCha20 in place of AES

The change anticipates the possible integration of the [ephemeral-fs 
](https://github.com/globaleaks/globaleaks-whistleblowing-software/tree/feature/ephemeral-fs)

### DIFF
--- a/backend/globaleaks/handlers/base.py
+++ b/backend/globaleaks/handlers/base.py
@@ -364,8 +364,6 @@ class BaseHandler(object):
             if self.request.args[b'flowChunkNumber'][0] != self.request.args[b'flowTotalChunks'][0]:
                 return None
 
-            f.finalize_write()
-
         mime_type, _ = mimetypes.guess_type(self.request.args[b'flowFilename'][0].decode())
         if mime_type is None:
             mime_type = 'application/octet-stream'
--- a/backend/globaleaks/handlers/recipient/export.py
+++ b/backend/globaleaks/handlers/recipient/export.py
@@ -221,7 +221,6 @@ class ExportHandler(BaseHandler):
         with stf.open('w') as f:
             for x in zipstream:
                 f.write(x)
-            f.finalize_write()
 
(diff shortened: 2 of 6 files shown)


msmannan00:
@evilaliv3 everything is looking good so far just wanted to confirm 3 points

1. Do we actually need "seek" and "tell" particularly for the implementation as its not being used
2. You are at the moment using same key in case if same file name is given, wouldn't that be an issue 
3. I am not sure, is ram protection implementing in existing system because keys are at the moment saved in ram without flushing it

evilaliv3:
Thank you for your review @msmannan00 

I've addressed point 1 and 2.

Here the answers to your questions point by point.
1. You are right seek and tell will be needed only when merging https://github.com/globaleaks/globaleaks-whistleblowing-software/tree/feature/ephemeral-fs to possibly integrating clamav  and MAT2. I will proceed simplifying the code for the moment.
2. Actually the key is random. What you noticed is the usage of the same salt that i was considering to generate on the base of the random filename for a matter of optimization.  But actually we could bind it to the key that is always re-generated.
3. Actually this problem is more broad and applies to the entire software. It is a problem with no known . In fact In Pyhon Strings are immutable; One could create a mutable object where to store a copy of the key and overwrite the copy, but still there is no way to overwrite original area of memory that contains the initial results of os.urandom(). As well when propagating the key to a function (e.g. to the encryption and decryption function) there will be automatic copies of the keys made depending on the libraries used and sill there wont be any way to ensure those copies are deleted. The same applies in general to any area of memory that contains plaintext information (e.g. HTTP Headers that contains keys, plaintext information before encryption, plaintext information after encryption); The typical countermeasures are to use mlock and avoid swap.

msmannan00:
Got it, looking forward towards the merge, anyhow really interesting update

PR #4375, opened 2025-01-12
Feature/client password hashing

evilaliv3:
This pull requests moves password hashing on with the following properties:
- Key derivation is now performed on the client with 16 steps and memory requirement set to 128MB
- The backend will preserve a sha512 hash of the key in order to perform authentication
- Users with a previous configuration will be able to perform a first login supported by the backend and automatically will transition into client hashing mode after their first password change
- Whistleblowers of an existent system with existing will continue to need to be supported by the backend with backend key derivation for old and new report; New systems, systems with no reports or systems where every existing reports expire will automatically start using client key derivation
- proof of work has been re-implemented replacing sha256 with argon2id with 1 iteration and 1MB of RAM requirement;
- python base64 encoding has been replaced with nacl.encoding.Base64Encoder preferring usage of functions designed in the context of cryptographic applications


evilaliv3:
@msmannan00 : when you like please feel free to proceed with the review of this pull request as well! thank you

msmannan00 (approved):
Overall, the code looks well-structured and functional. However, a few minor
improvements could enhance its readability:

The following code has really long lengthy lines that can be moved to seperate function but its just a suggestion, below we can make a separate function that returns these three keys below
key = Base64Encoder.decode(GCE.argon2id(receipt.encode(),
salt, GCE.options['OPSLIMIT'] + 1, 1 << GCE.options['MEMLIMIT']))
hash = GCE.argon2id(receipt, salt, GCE.options['OPSLIMIT'], 1 <<
GCE.options['MEMLIMIT'])

2. their are some comment, I guess left from my previous implementation that should be removed

3. we are using "hello world" in test cases maybe we should use something like GL_TEST instead of using such sample strings

evilaliv3:
Thank you @msmannan00 

Will work on your suggestions!

evilaliv3:
@msmannan00 : i should have applied most of your suggestions.

I've not found the comments that you think are old and should be removed; would you please point me to the exact lines of code?

PR #4380, opened 2025-01-24
index issue fixes in process e2e test

msmannan00:
minor index issue was fixed that was causing code crash


msmannan00:
@evilaliv3  small index bug in e2e test fixed

evilaliv3:
Super! Thank you @msmannan00 

Let me proceed mergin right away!

PR #4384, opened 2025-01-29
[Feature] version property exposed on public_node

marchrius:
This PR let a client retrieve the globaleaks version also from the `/api/public` endpoint

Before submitting a pull request, please ensure the following:

- [x] The pull request includes a description of the problem you're trying to solve.
- [x] The pull request provides an overview of the suggested solution.
- [x] The proposed code is fully functional.
- [x] The proposed code includes relevant tests to verify its functionality.
- [x] All new and existing tests pass successfully.
- [x] Overall code quality and test coverage metrics are not reduced by more than 0.5%

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,3 +1,12 @@
+Changes in version 5.0.44
+    Move password hashing on clientside
+    Upgrade proof of work from sha256 to argon2id with 1 step and 1MB of requirement
+    Refactor SecureTemporaryFile replacing AES with ChaCha20
+    Enable Brotli compression of static (offline) and dynamic (on-the-fly) content
+    Fix bug visualization of files uploaded by recipients introduced in 5.0.19
+    Bump client dependencies to latest version
+    Update translations
+
 Changes in version 5.0.43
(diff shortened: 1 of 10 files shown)


evilaliv3:
Thank you @marchrius for this proposal.

May i ask you which is hte motivation of this implementation?

The reason why we are not exposing the the software version is the following.

Back in 2011 we used to expose the software version and a auditor advised to remove it.

We do not believe in security through obscurity and of course the version of the software can be discovered with fingerprinting technics but still we want to honor and respect the peer review received when lecit :)

marchrius:
Hi @evilaliv3 , the explanation is simple. We have a small Globaleaks SDK which is implemented in our software. We need to know which version is running and compare it with remote available tags.

evilaliv3:
Thank you @marchrius for the clarification.

I think for exposing a detail like this, that is sensible to security we might eventually need an other strategy  and administrative API functionality because automating deployments does not justify lowering security.

If you like, please join us on community.globaleaks.org and if helpful you are invited to present your SDK in a next community call.

evilaliv3:
Closing as not in line with security constraints.

Issue #4540, opened 2025-06-26
Questions 'hints' do not appear on questions groups


evilaliv3:
As spotted by @susannaferro apparently questions 'hints' seems to not appear on questions group.

I can confirm this is the case and probably the bug exists since initial version 5.0.0

PR #4541, opened 2025-06-26
Fix issue #4540

evilaliv3:
This should fix the issue #4540 reported by @susannaferro 

The popover was actually not implemented. 

I consider that we could ship this in next minor 5.0.72


evilaliv3:
@msmannan00 : would you please check this pull request and proceed with your review?

Is all is correct you should be able to mark the pull request as reviewed and merge it.

PR #4565, opened 2025-07-02
Feature: Implement Docker-based building and client tests

AnXh3L0:
This PR aims to improve the current Docker-building methodology, by re-using the same build.yml file while adding a new install-docker script. Among the changes, we have implemented parallel Docker building and test execution on all the distros we currently support, with bookworm as the default one.

Additionally, the Cypress tests have been changed to run on these Docker instances, and Lighthouse testing has been added for the three main public-facing pages, that do not require log in.

Currently, the workflows may fail because there's a few Cypress tests failing, but the workflow is generally stable. A few changes have been introduced to the .dockerignore and the .gitignore files, in order to prevent local tests from being accidentally commited to the remote repo.

Moreover, almost every step of the build is commented, in order to make it easier to understand what each line does. At the moment, the workflow takes about 30-40 minutes to execute, when running the default bookworm build process.


evilaliv3:
Hey @AnXh3L0 

I just noticed this now, shall i proceed with a first round of review or is there still something you want to complete?

I will give this a try tomorrow.

I see that the test run is failing but it is probably due to problems related to the devel branch, while for this changes we could directly instegrate inside the stable.

AnXh3L0:
@evilaliv3 yes, I think you can start with a first round of review, I don't think there's anything else I can add for now.

evilaliv3:
I remodeled the pull request to be based on current stable branch in order to perform a first review.

.gitignore
```diff
@@ -3,3 +3,20 @@ backend/workingdir/
...
+docker-build/
+
+# Temporary build directories (used by enhanced test-docker.sh)
+../temp-build-*/
+
+# Build artifacts
+build/
+
+# Lighthouse reports generated by CI/CD
+client/lighthouse-results/
```
evilaliv3: As discussed let's remove thesechanges will we figure out in a separated ticket how to handler .gitignore

docker/.dockerignore
```diff
@@ -0,0 +1,14 @@
...
+!deb/
+!install-docker.sh
+!Dockerfile*
+
+# Exclude any backup or temporary files
+*.bak
+*.tmp
+*.log
+.DS_Store
+Thumbs.db
```
evilaliv3: For the moment lets skipp to add a dockerignore file trying to understand better how to handle it. Anyway for example on github there won't be files like .DS_Store or Thumds.db

docker/Dockerfile
```diff
@@ -1,16 +1,59 @@
...
+
+# Run the appropriate installation method based on INSTALL_FROM argument
+RUN echo "DEBUG: INSTALL_FROM=${INSTALL_FROM}, DISTRIBUTION=${DISTRIBUTION}" && \
+    if [ "${INSTALL_FROM}" = "local" ]; then \
+        echo "--- Performing local installation ---"; \
+        chmod +x /tmp/install-docker.sh && \
+        /tmp/install-docker.sh; \
+    else \
+        echo "--- Performing remote installation ---"; \
+        wget https://deb.globaleaks.org/install-globaleaks.sh -O /tmp/install-globaleaks.sh && \
```
evilaliv3: it was discussed to side load the file locally without downloading it from deb.globaleaks.org

docker/docker-compose.yml
```diff
@@ -1,16 +1,147 @@
...
+      interval: 10s
+      timeout: 5s
+      retries: 12
+      start_period: 30s
+    environment:
+      - DISTRIBUTION=focal
+    profiles:
+      - testing
+      - test-focal
+      - test-all
```
evilaliv3: this seems very doplicate once for each distro/version but if really necessry fine to keep this for now

docker/install-docker.sh
```diff
@@ -0,0 +1,95 @@
+#!/bin/bash
+# Docker-specific GlobaLeaks installation script
```
evilaliv3: Which are the specific details of this script? It appears mostly duplicated.

I suggest to evaluate to amend the original script with the few changes that are necessary to make it work in both the contexts.

evilaliv3 (changes requested):
Hello @AnXh3L0 , i've loaded all my current comments suggesting the changes to apply before merge.

AnXh3L0:
Thanks for the comments @evilaliv3, I am checking and applying the changes you requested.

docker/docker-compose.yml
AnXh3L0: These are necessary if we want to be able to run multiple/parallel Docker tests for the different distros/versions

docker/docker-compose.yml
```diff
@@ -1,16 +1,147 @@
...
     volumes:
       - globaleaks:/var/globaleaks
     ports:
       - 80:8080
+      - 8080:8080
       - 443:8443
+      - 8443:8443
+    profiles:
+      - production
+      - ""  # Default profile
```
evilaliv3: @AnXh3L0: which is the difference in these two profiles and when are they used?

docker/Dockerfile
```diff
@@ -1,14 +1,58 @@
-FROM debian:bookworm-slim@sha256:36e591f228bb9b99348f584e83f16e012c33ba5cad44ef5981a1d7c0a93eca22
+# GlobaLeaks Unified Dockerfile
+# Supports both production (remote) and testing (local) installs.
+#
+# Build Arguments:
+#   BASE_IMAGE:     The base image for the OS (e.g., debian:bookworm-slim)
+#   DISTRIBUTION:   The target distribution codename (e.g., bookworm)
+#   INSTALL_FROM:   Installation mode. 'remote' (default) or 'local'.
 
+ARG BASE_IMAGE=debian:bookworm-slim
```
evilaliv3: for reproducibility reason, i suggest to keep the sha256 of the current latest image

docker/docker-compose.yml
```diff
@@ -1,16 +1,147 @@
...
+
+  # Testing services for each distribution
+  globaleaks-test-bookworm:
+    build:
+      context: ..
+      dockerfile: docker/Dockerfile
+      args:
+        INSTALL_FROM: local
+        DISTRIBUTION: bookworm
+        BASE_IMAGE: ${BASE_IMAGE_BOOKWORM:-debian:bookworm-slim}
```
evilaliv3: @AnXh3L0: could we add the hash in any BASE_IMAGE defined in this file?

scripts/install.sh
```diff
@@ -161,18 +218,56 @@ if [ -d /globaleaks/deb ]; then
...
-    DO "apt-get install globaleaks -y"
+    if [ $DOCKER_ENV -eq 1 ]; then
+      DO "apt-get install globaleaks -y -o Dpkg::Options::=\"--force-confdef\" -o Dpkg::Options::=\"--force-confold\""
+    else
+      DO "apt-get install globaleaks -y"
+    fi
+  fi
+  
+  # For remote installation, also prevent service restart in containers
+  if [ $DOCKER_ENV -eq 0 ] && ([ -f /.dockerenv ] || [ -n "${container}" ] || grep -q '/docker\|/lxc' /proc/1/cgroup 2>/dev/null); then
```
evilaliv3: @AnXh3L0 : would you please explain to me this line and what it makes? when do we need lxc?

evilaliv3 (changes requested):
Thank you @AnXh3L0 

I think all is good to go. I would amiliorate just adding the hashes of all the base images used

docker/Dockerfile
AnXh3L0: Ok, adding back the sha256

scripts/install.sh
AnXh3L0: It checks if we are running inside a container, in case the first time we don't get a correct detection, sort of like a failsafe. The lxc is actually not necessary in this case, since I didn't know if any of our users ever deployed GlobaLeaks in lxc, so I assume we can probably remove it completely and only check for Docker instead.

evilaliv3: yes i think so, better to remove something that we are not controlling or that was explicitly requested by a user actively using it.

docker/docker-compose.yml
AnXh3L0: It's just for clarity/declaration purposes, it doesn't really do much else. Basically it allows us to test for every distro/case with a single command, and if necessary, it can be re-used in the future to run different things when the specific profile is chosen.

evilaliv3:
Thank you @AnXh3L0 

Actually i notice still duplication in hash definition and hash usage.

As example i refer to:
```bash
BASE_IMAGE_BOOKWORM: debian:bookworm-slim@sha256:36e591f228bb9b99348f584e83f16e012c33ba5cad44ef5981a1d7c0a93eca22
```

and:
```bash
BASE_IMAGE: ${BASE_IMAGE_BOOKWORM:-debian:bookworm-slim@sha256:36e591f228bb9b99348f584e83f16e012c33ba5cad44ef5981a1d7c0a93eca22}
```

Above is defined an hash but while using it the hash is provided again. Is this necessary or could we write it once?

evilaliv3 (approved):
🚀 LGTM @AnXh3L0 !

In my review i've applied the following changes and attempted to reduce code duplication:
    - restored original static docker-compose.yml used for production purposes
    - extended original docker-compose with health checks as in new testing compose
    - add missing testing of debian bullseye
    - removed usage of deprecated apt-key

Proceeding with the merge!

Issue #4566, opened 2025-07-02
Limit subdomain length


evilaliv3:
This ticket is to keep track of the activities related to limiting the subdomain length for tenants.

the fix is necessary both for reasons of functionality and accessibility since too long names either not work technically or from usability point of view.

evilaliv3:
with the previous patch i'm proposing to lower the limit subdomains to 40 that is definitely more than enough for an accessible name.

We might consider testing this length and eventually shorten this in the future.

PR #4575, opened 2025-07-03
Fix issue #4566

evilaliv3:
This should fix the issue https://github.com/globaleaks/globaleaks-whistleblowing-software/issues/4566

I consider that we could ship this in next minor 5.0.72

### DIFF
--- a/client/app/src/pages/admin/sites/siteslist/siteslist.component.html
+++ b/client/app/src/pages/admin/sites/siteslist/siteslist.component.html
@@ -60,7 +60,7 @@
           </div>
           <div class="form-group">
             <label>{{'Subdomain'|translate}}</label>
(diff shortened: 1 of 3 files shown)


evilaliv3:
@msmannan00 : would you please check this pull request and proceed with your review?

Is all is correct you should be able to mark the pull request as reviewed and merge it.

msmannan00 (approved):
The changes are pretty straightforward. Just one thing—could this impact the existing system? For example, users who already have subdomains exceeding the new length limit might encounter issues when making minor updates. They could suddenly find that 40-character subdomains are no longer supported and be forced to rethink and shorten them. That said, 40 characters is already quite long, so this probably won’t be an issue for most users

evilaliv3:
Thank you @msmannan00 for your wise consideration.

No it wont impact users since it will only apply to new configurations or edited configurations.

Issue #4581, opened 2025-07-09
Date field not showing correctly in a different language.


mdio33:
### Prerequisites

- [x] I have read the [contributing guidelines](https://github.com/globaleaks/globaleaks-whistleblowing-software/blob/main/.github/CONTRIBUTING.md)
- [x] I have [searched existing issues](https://github.com/globaleaks/globaleaks-whistleblowing-software/issues?utf8=%E2%9C%93&q=is%3Aissue) for duplicates or prior discussions
- [x] I have attempted to [reproduce the issue](https://try.globaleaks.org) on [try.globaleaks.org](https://try.globaleaks.org)

### What version of GlobaLeaks are you using?

Latest version and in try.globaleaks.org

### On which distribution/version is GlobaLeaks installed?

Debian 24.04.2

### What browser(s) are you seeing the problem on?

All

### What operating system(s) are you seeing the problem on?

Windows

### Describe the issue

Hello.

<img width="431" height="701" alt="Image" src="https://github.com/user-attachments/assets/e90e570e-0377-46b9-83d1-bb796dad305c" />

Using the latest version (updated yesterday) and on your test system, when trying to use a date field with the Portuguese language, there is overlapping of characters as you can see in the image.

Thank you for your help.

AnXh3L0:
Thanks for opening an issue for this @mdio33. We are currently triaging and will let you know soon.

AnXh3L0:
Thanks for the wait @mdio33, we have successfully triaged the issue and added the Bug label. We have identified the root cause and will release a fix for it in the upcoming 5.0.73 release!

mdio33:
Thank you.

I don't know if that happen with other languages, but i'm sure you will keep that in mind. :)

Thanks again.

AnXh3L0:
@mdio33 this issue has been resolved in the latest merged pull request https://github.com/globaleaks/globaleaks-whistleblowing-software/pull/4585. It will be included in the upcoming 5.0.73 release.

PR #4585, opened 2025-07-14
Fix for issue #4581

AnXh3L0:
In this pull request, I have addressed the issue https://github.com/globaleaks/globaleaks-whistleblowing-software/issues/4581 and implemented the following changes:

- Registered missing locales/languages from the locale-provider.ts helper file, fixing Angular error NG0701.
- Changed formatDate function in the custom-datepicker-i18n.ts file from the previous EEE to the EEEEE format, in order to display the short version for the week days, preventing the text overflow in several languages/locales with longer strings. This solution also prevents the need for visual/CSS changes.

### DIFF
--- a/client/app/src/shared/services/custom-datepicker-i18n.ts
+++ b/client/app/src/shared/services/custom-datepicker-i18n.ts
@@ -34,7 +34,7 @@ export class CustomDatepickerI18n extends NgbDatepickerI18n {
                                             // Ora e sempre resistenza!
     const targetDate = new Date(baseDate);
     targetDate.setDate(baseDate.getDate() + (weekday - 3));
-    return formatDate(targetDate, 'EEE', this.locale);
+    return formatDate(targetDate, 'EEEEE', this.locale);
   }
 
   // Full month name using Angular's formatDate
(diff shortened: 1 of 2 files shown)


client/app/src/shared/services/custom-datepicker-i18n.ts
```diff
@@ -15,7 +15,7 @@ export class CustomDatepickerI18n extends NgbDatepickerI18n {
     this.translationService.currentLocale$.subscribe(locale => {
       if (locale === 'crh') {
         this.locale = 'tk';
-      } else if (locale === 'dv') {
+      } else if (locale === 'ba' || locale === 'dv' || locale.startsWith('ug')) {
```
evilaliv3, line 18: As noted there are other languages that requires specific handling like: "ba", "ug", "ug@cyrl", "ug@latin"

client/app/src/shared/services/custom-datepicker-i18n.ts
```diff
@@ -34,7 +34,7 @@ export class CustomDatepickerI18n extends NgbDatepickerI18n {
                                             // Ora e sempre resistenza!
     const targetDate = new Date(baseDate);
     targetDate.setDate(baseDate.getDate() + (weekday - 3));
-    return formatDate(targetDate, 'EEE', this.locale);
+    return formatDate(targetDate, 'EEEEEE', this.locale);
```
evilaliv3, line 37: As discussed probably "EEEEEE" seems to work on every language and being longer than "EEEEE" probably might offer better usability.

evilaliv3 (approved):

Issue #4214, opened 2024-10-03
Extend audit log to enable to filter order columumns


evilaliv3:
### Proposal

This ticket is to track the implementation of some extentions to audit logs to empower administrators to perform auditing activities.

### Motivation and context

Many users in recent times had the need to try to reconstruct:
- what happened on a certain report
- how a users has operated

The current platform logs almost that it is needed but this information is not easy accessible.
For example:
- it is not possible to filter for a specific report
- it is not possible to filter for a specific user
- it is not possible to filter for a specific log type
- it is not possible to filter for a certain interval of time
- it is not possible to do a free text search on the events

In addition the information is not so accessible (users and reports are identified by ID instead of name while  we could possibly export also the username and the report number)

JaviAlama:
Hello colleagues. Everything indicated in this ticket can be done very easily from the log files provided by the platform.

In our case, using Excel, we can do everything indicated with 1-2 clicks.

For us, although interesting, these are not priority functionalities right now.

evilaliv3:
Thank you @JaviAlama for your feedback. I really appreciate.

For sure it is not priority and I consider the same from the point of view of auditing that definitely can be done with excel. Is from the security perspective that i think we should support users to be able to not export audits logs if not necessary because of course by doing this some metadata is exported, about timing of reports, about status of reports, about timing of access by users that can expose whistleblowers if put together with other information.

msmannan00:
@evilaliv3 starting to work on this ticket

evilaliv3:
Perfect thank you! this is definitely one that benefits of some help.

It is strictly related to https://github.com/globaleaks/globaleaks-whistleblowing-software/issues/4085 so we might try to see the similarity and goes in the same direction of @AnXh3L0

evilaliv3:
Thank you @AnXh3L0, i should have performed the review of this changes.

<img width="1920" height="1940" alt="Image" src="https://github.com/user-attachments/assets/3047bd9d-79ec-412d-b3e3-d76611199025" />

Within the integration i took care of applying same layout to:
- applied some refactor for reasons of performance generating optimized processing datastructures for the processing of the logs
- extended the log entries descriptors with all the types of logs managed
- aligned existing search bars within the application (on reports list and site list)
- aligned existing sort interfaces (on reports list)

Since this changes are now into devel and our goals for the ticket at addressed i consider we could close the ticket.

PR #4593, opened 2025-07-29
Added table sorting, searching and filtering for audit log table

AnXh3L0:
I have applied sorting for all the table columns, and filtering for the date and type columns, re-using elements from other parts of the code to apply them here as well. A search bar has been added, which allows the users to search for anything on all columns apart from the date one.

Colored dots have been added for the log types, and they have been categorized into 3 different categories now, from access, update and delete.

Users column has been changed to now show the full user name and their username in parentheses.

These changes cover the #4214 requirements.


evilaliv3:
Closing as merged manually into devel.

Main changes applied:
- Reorganize audit log filters separeting categories in lists optimizing code and moving functions to Utils service
- Ensure icons are included using uniform code format: '<i class="fas fa-icon"></i>'
- Remove inclusion of not existing CSS
- Style linting

PR #4654, opened 2025-10-07
Refactor access_report grouping logic in audit log

AnXh3L0:
Changed grouping to single user, sequential, same-action, same-day only


evilaliv3:
From the meeting, i discovered i might have not integrated some parts specifically the filter on the sure. Let me pick up this

PR #4747, opened 2026-02-15
Add path validation and input format checks for password reset tokens

chiranjeevi-max:
Before submitting a pull request, please ensure the following:

- [x] The pull request includes a description of the problem you're trying to solve.
- [x] The pull request provides an overview of the suggested solution.
- [x] The proposed code is fully functional.
- [x] The proposed code includes relevant tests to verify its functionality.
- [x] All new and existing tests pass successfully.
- [x] Overall code quality and test coverage metrics are not reduced by more than 0.5%

## Problem

The password reset functionality uses user-supplied `reset_token` values directly in file path operations without validation. This could allow malformed input to reference unintended file paths.

## Solution

Add two layers of input validation:

1. **Input format validation:** Added `token_regexp` pattern (`^[a-f0-9]{64}$`) to validate reset tokens match the expected 64-character hexadecimal format before processing

2. **Path validation:** Added `directory_traversal_check()` calls before file operations in:
    * `validate_password_reset()` in `reset_password.py`
    * `change_password()` in `operation.py`

## Changes

* `backend/globaleaks/rest/requests.py` - Added token format regex validation
* `backend/globaleaks/handlers/user/reset_password.py` - Added path validation
* `backend/globaleaks/handlers/user/operation.py` - Added path validation
* `backend/globaleaks/tests/handlers/user/test_reset_password.py` - Updated tests for new validation

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -1,18 +1,20 @@
 name: Report a bug
-description: Tell us about a bug or issue you may have identified.
-title: "Provide a general summary of the issue"
+description: Tell us about a bug or issue you have identified.
+title: "Bug: [Short summary of the issue]"
+
 body:
-  - type: checkboxes
+  - type: markdown
     attributes:
-      label: Prerequisites
-      description: Help us resolve issues faster by completing the following.
-      options:
-        - label: I have read the [contributing guidelines](https://github.com/globaleaks/globaleaks-whistleblowing-software/blob/main/.github/CONTRIBUTING.md)
-          required: true
-        - label: I have [searched existing issues](https://github.com/globaleaks/globaleaks-whistleblowing-software/issues?utf8=%E2%9C%93&q=is%3Aissue) for duplicates or prior discussions
-          required: true
-        - label: I have attempted to [reproduce the issue](https://try.globaleaks.org) on [try.globaleaks.org](https://try.globaleaks.org)
-          required: true
+      value: |
+        ## Guidance
+
+        Please use this template to clearly describe the reported bug. Well-structured reports help us identify and fix issues efficiently.
+
+        Before proceeding, please ensure you:
+        - Review the [Contributing Guidelines](https://github.com/globaleaks/globaleaks-whistleblowing-software/blob/main/.github/CONTRIBUTING.md)
+        - Follow the project [Code of Conduct](https://github.com/globaleaks/globaleaks-whistleblowing-software/blob/stable/CODE_OF_CONDUCT.md)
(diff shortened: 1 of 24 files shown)


chiranjeevi-max:
Hi @evilaliv3 ,

can you review this change?

Also, I noticed a CI failure in `18-test-globaleaks-process.cy.ts` (timeout waiting for #TipCommentsBox), which appears to be not related to my change. The password reset test (`04-test-user-password-reset.cy.ts`) passes.

evilaliv3:
Thank you @chiranjeevi-max 

Do you mind clarifying the attack possibilities that you have identified and what lead you to discover the situation?

Your rational is legit but we typically try to avoid adding code that is just reduntant since every line require maintainance. Do you get what i mean?

chiranjeevi-max:
Hi @evilaliv3, yes, I understand. I am happy to clarify on this change.

**Attack scenario:**
In `validate_password_reset()`, the `reset_token` parameter from the HTTP request is passed directly to `os.path.join()` and `open()`:
```
with open(os.path.abspath(os.path.join(State.settings.ramdisk_path, reset_token)), "r") as f:
    token = f.read()
```
An attacker can supply `reset_token` = "../../../etc/passwd" (or an absolute path like `/etc/passwd` which `os.path.join` uses directly). 

This allows:
* **Arbitrary file read** - The file content is read into `token`, then split by : to extract `user_id`. While the data isn't returned directly, timing and error differences can leak information.
* **File existence oracle** - Different exceptions for existing vs non-existing files.

**Why this matter here?**
The `PUT /api/user/reset/password` endpoint has `check_roles = 'any'`, meaning it's **unauthenticated**. Given GlobaLeaks purpose as whistleblowing software where source protection and confidentiality are critical, an unauthenticated file read/existence oracle could be used to probe system configuration, enumerate the environment, or gather information for further attacks.

**Why not redundant:**

The regex validation (`^[a-f0-9]{64}$`) rejects malformed input at the API boundary before it reaches filesystem operations. Valid tokens from `generateRandomKey()` are always 64 hex characters, so this adds no overhead for legitimate requests. Only malicious ones are rejected early.

The `directory_traversal_check()` provides defense-in-depth consistent with `StaticFileHandler` and `FileHandler`.

evilaliv3:
Thank you, very correct analysis and valid.

will review and integrate the patch shortly.

I'm curious are you currently auditing the software or implementing it for some projects? Please feel free to reach out to me directly on community.globaleaks.org

chiranjeevi-max:
Thanks @evilaliv3! 

I'm not currently implementing it anywhere. I'm just interested in contributing to the project. 

Whistleblower protection is an important cause 💝, and I wanted to help where I could.

Happy to continue looking at the codebase if there are other areas that could use review.

PR #4753, opened 2026-02-20
Fix/docker amd64 platform

### LINKED ISSUE #4751: Bug: Docker image published as i386 (32-bit) causes OverflowError on /api/recipient/rtips — all recipient logins fail

5.0.88 Docker Image (but all Docker Hub tags are affected probably)

### Environment
- AWS ECS (EC2 launch type), instance t3a.small — x86_64, 64-bit host
- Docker image: `globaleaks/globaleaks:latest` pulled from Docker Hub
- Region: eu-west-1

---

## Symptom

All **recipient** accounts fail to log in with a generic validation error.
The **admin** account works normally.

The failing API call is:
```
GET /api/recipient/rtips → HTTP 500
```

Traceback received via error notification email:
```
File "/usr/lib/python3/dist-packages/globaleaks/handlers/recipient/__init__.py", line 33
    updated_before = datetime.fromtimestamp(int(args.get(b'updated_before', [b'32503680000'])[0]))
OverflowError: timestamp out of range for platform time_t
```

---

## Root cause — confirmed

The Docker image published on Docker Hub is built for **i386 (32-bit)**, not amd64.

Diagnostic evidence collected inside the running container:

| Check | Result |
|---|---|
| `docker inspect` → Architecture | `386` |
| `uname -m` inside container | `x86_64` (reflects host kernel, not userspace) |
| `python3 -c "import struct; print(struct.calcsize('P')*8)"` | `32` |
| `python3 -c "from datetime import datetime; datetime.fromtimestamp(32503680000)"` | `OverflowError` |

**The host EC2 instance is correctly 64-bit.** The problem is entirely in the published Docker image.

Because the container userspace is i386, Python is compiled as 32-bit. The `time_t` type is therefore limited to a maximum value of `2147483647` (year 2038). The hardcoded default timestamp `32503680000` (~year 3000) used in `get_receivertips()` exceeds this limit and raises `OverflowError`.

The admin account is unaffected because it uses different API endpoints (`/api/admin/`) that do not call `get_receivertips()`.

---

## Why the Docker image ends up as i386

The `Dockerfile` in the repository (`docker/Dockerfile`) does not specify a `--platform` directive on the `FROM` instruction. Without an explicit `FROM --platform=linux/amd64`, Docker Hub's build pipeline defaults to `i386` on some builder configurations, and the resulting manifest is tagged as `386`.

This is reproducible for **every version currently published on Docker Hub** — there is no tag that resolves to `amd64`.

---

## What the changelog says — and what it does not fix

The changelog entry for version 5.0.x reads:

> `[doc] Revise documentation informing users that a 64-bit system is required to run GlobaLeaks (#4202)`

This is a documentation-only change. It does not fix the Dockerfile, does not fix the default timestamp value, and does not fix the published image. Users deploying via Docker (the documented method) have no way to obtain a working 64-bit image from Docker Hub today.

## Requested fix

Add `--platform linux/amd64` to the `FROM` instruction in `docker/Dockerfile` and rebuild + republish all supported tags:
```dockerfile
FROM --platform=linux/amd64 debian:trixie-slim
```

And ensure the CI/CD pipeline that publishes to Docker Hub explicitly passes `--platform linux/amd64` (or `linux/amd64,linux/arm64` for multi-arch).

Verification:
```bash
docker inspect globaleaks-amd64:local | grep Architecture
# Expected: "Architecture": "amd64"

docker run --rm globaleaks-amd64:local python3 -c \
  "import struct; print(struct.calcsize('P')*8, 'bit')"
# Expected: 64 bit
```

leofalp:
## Problem

The Docker image published on Docker Hub resolves to **i386** when no platform is explicitly specified during build. This means that on a 64-bit host, the container userspace still runs **32-bit Python**, where `time_t` is a signed 32-bit integer.

The default timestamp `32503680000` (~year 3000) used in `get_receivertips()` overflows this limit, causing:

```
OverflowError on /api/recipient/rtips
```

This breaks **all recipient logins** for users running the official Docker image.

---

### 1. Missing `--platform` in Dockerfile
The `FROM` instruction had no `--platform` directive, leaving Docker free to resolve the base image to the host's default variant — which in some environments (e.g. multi-arch manifests) resolves to `i386`.

### 2. `generate_docker_compose.sh` was not passing build args
The generated `docker-compose.yml` used `build: .` (shorthand), which means **neither `BASE_IMAGE` nor `DISTRIBUTION` were passed as build-time arguments**. As a result:
- The Dockerfile always used its hardcoded `ARG` defaults, silently ignoring the distro selected at runtime.
- The `DISTRIBUTION` variable was only injected as a runtime environment variable, not as a build arg where the Dockerfile actually needs it.

---

### `docker/Dockerfile`
Added `--platform=linux/amd64` to the `FROM` instruction to explicitly target the correct architecture:

```dockerfile
FROM --platform=linux/amd64 ${BASE_IMAGE}
```

### `docker/generate_docker_compose.sh`
Replaced the `build: .` shorthand with the full `build:` syntax, passing both `BASE_IMAGE` and `DISTRIBUTION` as proper build arguments:

```shell
    build:
      context: .
      args:
        BASE_IMAGE: ${base_image}
        DISTRIBUTION: ${name}
```

---

## Note for maintainers

If the CI pipeline that **publishes to Docker Hub** uses `docker buildx` or `docker/build-push-action`, it should also explicitly set:

```yaml
platforms: linux/amd64,linux/arm64
```

to ensure the published manifest is correct for all supported architectures. This is outside the scope of this PR as it requires access to Docker Hub secrets, but it is the complementary fix needed on the CI side.

---

Fixes #4751

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -7,7 +7,7 @@
 ARG BASE_IMAGE=debian:trixie-slim@sha256:38af8d51386e2fc12cc71c038c6343205450691f739fc5840dad7380a31e382e
 ARG DISTRIBUTION=trixie
 
-FROM ${BASE_IMAGE}
+FROM --platform=linux/amd64 ${BASE_IMAGE}
 
 ADD https://deb.globaleaks.org/install.sh /tmp/install.sh
 
--- a/docker/Readme.md
+++ b/docker/Readme.md
@@ -3,6 +3,8 @@
 To run GlobaLeaks using Docker
 ```bash
  docker run -d --name globaleaks \
-  -p 80:80 -p 443:443 \
+  --platform linux/amd64 \
+  -p 80:8080 \
+  -p 443:8443 \
   -v globaleaks-data:/var/globaleaks \
   globaleaks/globaleaks:latest
--- a/docker/generate_docker_compose.sh
+++ b/docker/generate_docker_compose.sh
@@ -44,7 +44,11 @@ services:
   globaleaks:
     environment:
       DISTRIBUTION: ${name}
-    build: .
+    build:
+      context: .
+      args:
+        BASE_IMAGE: ${base_image}
+        DISTRIBUTION: ${name}
     restart: unless-stopped
     container_name: globaleaks
     network_mode: bridge


evilaliv3 (approved):
Very good fix and very clear pull request @leofalp , thank you!

Proceeding with the merge directly in stable branch and staging release 5.0.89

PR #4805, opened 2026-04-04
Security: Path traversal risk when loading profile by name

tuanaiseo:
## Problem

Profile path is built from `name` and directly joined into a filesystem path (`.../data/profiles/{name}.json`) without validation. If `name` is attacker-controlled in any call path, `../` sequences can escape the profiles directory and read unintended files.

**Severity**: `high`
**File**: `backend/globaleaks/models/profiles.py`

## Solution

Validate `name` against a strict allowlist (e.g., `^[a-zA-Z0-9_-]+$`), resolve with `os.path.abspath`, and enforce that the final path remains under the profiles directory before reading.

## Changes

- `backend/globaleaks/models/profiles.py` (modified)

## Testing

- [ ] Existing tests pass
- [ ] Manual review completed
- [ ] No new warnings/errors introduced

### DIFF
--- a/backend/globaleaks/models/profiles.py
+++ b/backend/globaleaks/models/profiles.py
@@ -13,7 +13,15 @@ def load_profile(session, tid, name):
     :param tid: The tenant id of the tenant to be configured
     :param name: The name of the profile to be used
     """
-    path = os.path.join(Settings.client_path, 'data/profiles', '{}.json'.format(name))
+    if not name or not all(c.isalnum() or c in ('_', '-') for c in name):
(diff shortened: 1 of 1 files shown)


evilaliv3:
Thank you for the suggestion @tuanaiseo !!

Your feedback is correct but the patch was a bit over complicated.
I just emitted an alternative patch that makes use of an existing directory_traversal_check utility.

Worth saying that the function load_profile was never used directly but only via the wizard.py handler that includes a regular expression check on the file name.

PR #4829, opened 2026-05-11
Fix channels and recipients image aspect ratios

casungo:
## Summary

This change preserves the natural aspect ratio of context and receiver images while keeping the existing maximum display bounds.

It also lets context images use the full card width, avoiding unnecessary compression for wide logos in the channel selection screen.

## Problem

Context images had a fixed rendered height combined with a maximum width. Very wide logos could therefore be squeezed horizontally instead of being scaled proportionally.

## Changes

- Use automatic image dimensions with max-width/max-height constraints for context, receiver, and upload preview images.
- Keep object-fit containment for bounded image previews.
- Remove the narrow column wrapper from context selection images.

## Validation

- Built the Angular client successfully with `npx ng build --configuration=development`.


evilaliv3:
Thank you for your contribution @casungo 

The patch looks good to me.

I notice that you are performing edits to both the context-image and receiver-image but the html is modified only fore the context.

Do you mind implement and test the changes also for the receiver image?

casungo:
Thank you for the review.

I updated the receiver image markup as requested, so receiver images now use the full card width consistently with context images.

While testing this on a real instance, I also refined the shared image bounds from `20rem` to `28rem` and updated the selected-context view to avoid keeping the context image constrained to a narrow column. The images now keep their natural aspect ratio while still being capped by `max-width` and `max-height`.

casungo:
The nested `@if (context.picture)` around the `` was removed because the parent block already guards the entire image row on the same condition. That makes the inner check redundant: Angular will only render the row and image markup when `context.picture` is true.

So the simplified version is functionally equivalent, and I’d keep the template with just the outer guard unless we want to preserve the nested block purely to minimize diff churn.

evilaliv3:
Very good patch thank you @casungo 

Tested and i'm going to merge it right away in release 5.0.93

evilaliv3:
Closing as directly integrated in 5.0.93

PR #4888, opened 2026-08-02
fix(backend): reject attachment uploads without session

### LINKED ISSUE #4782: AttributeError: 'NoneType' object has no attribute 'user_id'
### Discussed in https://github.com/orgs/globaleaks/discussions/4774

<div type='discussions-op-text'>

<sup>Originally posted by **isAAAc** March  6, 2026</sup>
Hello, 

We regularly encounter this error, which recurs repeatedly and then disappears either on its own or after restarting the globaleaks service.

From what we have been able to investigate and understand, this occurs when ‘certain’ attachments are added to a report, 
but it does not happen systematically.

Therefore, we wonder if this might ultimately be a bug in globaleaks.

Can you help us, please?

We are at your disposal if you need any further information.

OS:   `debian 12.13`
globaleaks version: `globaleaks 5.0.88`  (installed via apt)

error log:
```
Version: 5.0.88

AttributeError Attribute not found.

Traceback (most recent call last):

 File "/usr/lib/python3/dist-packages/twisted/internet/defer.py", line 190, in maybeDeferred
   result = f(*args, **kwargs)

 File "/usr/lib/python3/dist-packages/globaleaks/handlers/whistleblower/attachment.py", line 76, in post
   return register_ifile_on_db(self.request.tid, self.session.user_id, self.uploaded_file)

AttributeError: 'NoneType' object has no attribute 'user_id'
```

thx all for your answer ;)
</div>

PSR94:
Summary
- Add explicit authentication guards to whistleblower attachment upload handlers before accessing session fields.
- Add regression coverage for missing-session calls on both submission and post-submission attachment handlers.

Why this helps
- Prevents missing or expired sessions from surfacing as AttributeError server errors during attachment upload handling.
- Returns the existing NotAuthenticated API error for the invalid request state.

Changes made
- Guarded SubmissionAttachment.post and PostSubmissionAttachment.post when self.session is missing.
- Added focused Trial unit tests for both attachment handler paths.

Testing
- PYTHONPATH=/Users/pavan/Documents/Open\ Source\ Contributions/globaleaks-whistleblowing-software/backend env/bin/trial globaleaks.tests.handlers.whistleblower.test_attachment.TestAttachmentAuthentication
- env/bin/ruff check globaleaks/handlers/whistleblower/attachment.py globaleaks/tests/handlers/whistleblower/test_attachment.py

Related issue
Closes #4782


evilaliv3:
Thank you for your contribution @PSR94 !

Would you please clarify how have you experienced this issue and why you consider important adding this code?

PSR94:
Thanks for looking at it. I picked this up from #4782, where the reported traceback shows PostSubmissionAttachment.post reaching self.session.user_id while self.session is None during a whistleblower attachment upload.

I consider the guard useful because this is an invalid authentication/session state, so the endpoint should fail with the existing NotAuthenticated API error instead of surfacing an AttributeError as an internal server error. The added regression tests cover the missing-session path for both attachment handlers.

PR #4896, opened 2026-09-10
Log permanent attachment deletion in the audit log

### LINKED ISSUE #4894: Improve audit log tracking when a recipient delete a file

5.0.96

### Describe the issue

The recipient of the report can permanently delete a masked attachment

### Proposed solution (optional)

The recipient of the report can permanently delete a redacted attachment.
This is linked to the permission to permanently delete redacted information, but it is a highly impactful operation that is not even recorded in the system log. The log only shows the same message that appears when data is redacted, namely ‘update_redaction’.

vbrekher:
## Summary

When a recipient permanently redacts a masked attachment, the file is deleted but the audit log currently only records `update_redaction`.

This change adds a separate `delete_file` audit entry for that permanent deletion.

The new audit entry records:

- the recipient who performed the deletion
- the deleted file ID
- the associated report ID

The existing `update_redaction` audit entry is preserved.

This does not change deletion behavior, permissions, or redaction logic. It only makes the destructive file deletion explicitly visible in the audit trail.

## Testing

Added a regression test covering permanent deletion of a masked attachment.

The test verifies that:

- the file is permanently deleted
- exactly one `delete_file` audit entry is created
- the audit entry contains the expected user ID
- the audit entry identifies the deleted file
- the audit entry references the correct report

Fixes #4894

### DIFF
--- a/backend/globaleaks/handlers/recipient/rtip.py
+++ b/backend/globaleaks/handlers/recipient/rtip.py
@@ -1322,12 +1322,15 @@ def update_redaction(session, tid, user_id, redaction_id, redaction_data, tip_da
 
                 db_log(session, tid=tid, type='update_redaction', user_id=user_id, object_id=redaction.id, data=log_data)
 
+                file_id = redaction.reference_id
                 if session.query(models.ReceiverFile) \
-                          .filter(models.ReceiverFile.id == redaction.reference_id).count():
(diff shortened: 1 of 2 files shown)


evilaliv3:
Thank you @vbrekher !

We are performing some work on this matter in the devel branch.

We will review your pull request shortly!

